PR #7, opened 2023-04-06
PLT-5527 Reduced time to run JSON-RPC test scenarios in `marconi-sidechain` from a few minutes to a few seconds.

koslambrou:
* Refactored the test such that the JSON-RPC server is shared accross all tests.

* The `storableAction` called at each test makes sure that indexer is initialized at each call.


marconi-sidechain/test/Spec/Marconi/Sidechain/RpcClientAction.hs
```diff
@@ -0,0 +1,74 @@
...
+        (_ :<|> _ :<|> rpcSyncPoint :<|> rpcUtxos :<|> _ :<|> _ :<|> _)
+          = mkHoistedHttpRpcClient clientEnv
+    pure $ RpcClientAction (mkInsertUtxoEventsCallback env)
+                           (\a -> rpcUtxos $ TxOutAtQuery a Nothing)
+                           (rpcSyncPoint "")
+
+baseUrl :: Warp.Port -> BaseUrl
+baseUrl port = BaseUrl Http "localhost" port ""
+
+-- | A type for Storable Action to store events
```
kayvank, line 41: same as above. My error originally, but I think  `type`  should be changed to  `alias`

marconi-sidechain/test/Spec/Marconi/Sidechain/RpcClientAction.hs
```diff
@@ -0,0 +1,74 @@
...
+import Network.HTTP.Client (defaultManagerSettings, newManager)
+import Network.JsonRpc.Client.Types ()
+import Network.JsonRpc.Types (JsonRpcResponse)
+import Network.Wai.Handler.Warp (Port)
+import Network.Wai.Handler.Warp qualified as Warp
+import Servant.API ((:<|>) ((:<|>)))
+import Servant.Client (BaseUrl (BaseUrl), ClientEnv, HasClient (Client), Scheme (Http), client, hoistClient,
+                       mkClientEnv, runClientM)
+
+-- | Alias for the storable-action and RPC-client-Action pair, to simplify the type signatures
```
kayvank, line 21: My error originally, but I think  `Alias`  should be changed to  `Type`

kayvank (approved):
LGT. Response time is so much better. thank you !
What was the main fix?  Moving the `Warp.testWithApplication` from `TestGroup` to `main` ?

koslambrou:
> What was the main fix? Moving the Warp.testWithApplication from TestGroup to main ?

@kayvank That's right. Else, the `testWithApplication` is called for every single test cases generated by the property tests.

PR #14, opened 2023-04-14
PLT-595 Error handling - DRAFT

berewt:
Marconi-chain-index compiles

- SQL errors are explicitly caught.
- On error, a worker send a notification to the coordinator through an `MVar`, release its token, and continue its work.
- `Coordinator` check the error `MVar` before sending a new event.
- On error, a Coordinator throw an exception.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -450,3 +477,29 @@ runIndexers socketPath networkId cliChainPoint traceName list = do
...
+      Left err -> do
+          tryPutMVar errBox err $> (x, x)
+      Right x' -> pure (x', x')
+
+failWhenFull :: MonadIO m => MVar a -> m ()
+failWhenFull x = do
+  isEmpty <- liftIO $ isEmptyMVar x
+  if isEmpty
+  then pure ()
+  else error "an indexer raised an error"
```
berewt, line 505: I still don't like to exit workers with an error, but doing proper thread management is a bit of work: we need to gather the `ThreadId`, keep them in the `Coordinator` and use them to kill the threads on error. With the current design, it's another significant refactoring.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/AddressDatum.hs
```diff
@@ -95,7 +99,7 @@ data AddressDatumHandle = AddressDatumHandle
     , _addressDatumHandleDiskStore :: Int
     }
 
-type instance StorableMonad AddressDatumHandle = IO
+type instance StorableMonad AddressDatumHandle = ExceptT IndexerError IO
```
koslambrou, line 102: :( This definitely adds some boilerplate everywhere with `lift`.

And it reminds me off an (opinionated) article about the `ExceptT e IO a` anti-pattern.

I guess the question I have: how do we catch other IO exception that are not explicitly mentioned by `IndexerError`?

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -97,23 +103,27 @@ scriptDataFromCardanoTxBody _ = mempty
...
+  , _errorVar     :: !(MVar IndexerError)
   , _barrier      :: !QSemN
   , _indexerCount :: !Int
   }
 
+makeLenses 'Coordinator
+
 initialCoordinator :: Int -> IO Coordinator
-initialCoordinator indexerCount =
+initialCoordinator count =
```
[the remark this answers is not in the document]
berewt: I don't even remember having changed it, it may be an error during one of the rebase I've made while working on it, I'm reverting it.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/AddressDatum.hs
berewt: We don't, unless we try/catch them (which is exactly what I've done for the SQLite errors).
Changing signatures for `MonadIO m, MonadError IndexerError m` helps a bit, and that's what I've tried to do as much as I could, but we still have to `liftIO` function that lives in IO.

Not dealing with `ExceptT` and in the large, with both `MonadError` and `Control.Exception` is a motivation in many framework, the most common noe being `RIO` I guess.
It has other downside though, and usually comes with a strongly opinionated prelude.
The "new indexers" hide a bit of this complexity and gather almost all of it in its core, it's still messy, but less painful to handle in new indexers.

If you remember the discussion we had about error handling, that's why I said that being explicit would be a lot messier. I do agree that being explicit is more informative and avoid forgetting about error handling, but yes, it has a cost.

koslambrou: :+1: 

I guess in any `IO` code caller, he would need to wrap the runtime exception into our custom error type. I agree it's a trade-off. In the end, I think I'm okay with going this route if the rest of the team agrees.

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -274,24 +289,26 @@ epochStateWorker_
         nodeConfigPath
         onInsert
         securityParam
-        Coordinator{_barrier}
+        Coordinator{_barrier, _errorVar}
         ch
         dbPath = do
     nodeConfigE <- runExceptT $ readNetworkConfig (NetworkConfigFile nodeConfigPath)
-    nodeConfig <- either (error . show) pure nodeConfigE
+    nodeConfig <- either (throw . CantStartIndexer . Text.pack . show) pure nodeConfigE
```
koslambrou, line 296: So this throws a runtime error, right? It's not caught anywhere?

If it's thrown, the thread stops, but the application still continues to run right?

berewt: This one aren't using ExceptT yet, not sure if we want to introduce it here as well as we didn't start the indexers yet (we're about to start them).
This code is before any `forkIO` call, so it leaves in the "main" thread, and an uncaught exception will cause a failure of the main thread and thus of the whole application.

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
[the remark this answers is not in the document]
berewt: oh it was to avoid shadowing an existing binding. Switching it to `indexerCount'`

marconi-chain-index/src/Marconi/ChainIndex/Indexers/AddressDatum.hs
berewt: At least, if we want to allow the development of other backend aside the SQLite one, it's the safer way to do so: the coordinator can't just adapt its behaviour to any kind of implicit error sent along the `IO` monad with `throw`.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/AddressDatum.hs
```diff
@@ -245,8 +249,9 @@ instance Buffered AddressDatumHandle where
     :: Foldable f
     => f (StorableEvent AddressDatumHandle)
     -> AddressDatumHandle
-    -> IO AddressDatumHandle
-  persistToStorage es h = do
+    -> StorableMonad AddressDatumHandle AddressDatumHandle
```
eyeinsky, line 252: What if the wrapping to `EitherT e IO` happened within `Marconi.Core.Storable`, such that each and every indexer wouldn't need to change its `StorableMonad` to list it out explicitly?

berewt: It's the firt thing that I tried, but it implies roping SQLite in `core`, and was a nightmare to implement, I didn't keep track of why though.

berewt: Ok, I gave it another shot and I remember why I did not do it: it's doable, but you don't do any error management in Storable.
And yes we don't change StorableMonad but we have to modify all the signature that has a StorableMonad in it and that are related to the storable module. So to save one change, we impose many to the type signatures.

PR #15, opened 2023-04-15
PLT-5591 Including JSON Schema of expected request and response of marconi-sidechain endpoints.

koslambrou:
* Changed JSON-RPC types to reflect the new marconi-sidechain requirements.

* Renamed GetCurrentSyncedPointResult to GetCurrentSyncedBlockResult

* Created GetUtxosFromAddressParams for representing the inputs of the
  GetUtxosFromAddress method.

* Created GetUtxoFromAddressResult to replace AddressUtxoResult and
  changed representation of AddressUtxoResult to reflect a single UTXO.

* Created GetTxsBurningAssetIdParams 

* Replaced MintingPolicyHashResult with GetTxsBurningAssetIdResult

* Added JSON rountrip tests for
  * GetUtxoFromAddressParams
  * GetUtxoFromAddressResult
  * GetCurrentSyncedBlockResult
  * GetTxsBurningAssetIdParam
  * GetTxsBurningAssetIdResult

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


marconi-sidechain/README.md
```diff
@@ -198,147 +249,473 @@ As a consequence, if the last chainpoint of the utxo indexer can, at most,
...
+        "address": {
+          "type": "string",
+          "_comment": "Address encoded in the Bech32 format"
+        },
+        "createdAfterSlotNo": {
+          "_comment": "Not available yet",
+          "type": "integer"
+        },
+        "unspentBeforeSlotNo": {
+          "type": "integer"
```
berewt, line 373: Regarding the use of slot number alone instead of the whole chain point, are we sure that we can handle rollback in a safe way in this scenario?
I think it should be ok, but I didn't put many thoughts in it at the moment.
The main issue being that we may send inconsistent results for two different request if a rollback of 1 and a new push were done between two requests on the same slot.

berewt (approved):
LGTM, I filed an issue I have with the use of slot number in queries (it applies to other methods that use a slot number instead of a block info)

marconi-sidechain/README.md
koslambrou: In case of a rollback, we can't do anything about that imo. Inconsistent results should be handled by the user itself. How would changing the type of `unspentBeforeSlot` to a ChainPoint help the user?

PR #16, opened 2023-04-17
Compare Marconin's epoch nonces with cardano-db-sync (PLT-1395)

eyeinsky:
Compare results between Marconi and dbsync:
- [PLT-1395](https://input-output.atlassian.net/browse/PLT-1395): epoch->nonce
- [PLT-387](https://input-output.atlassian.net/browse/PLT-387): epoch->stakepool sizes

Because we don't run this test on CI (as it requires a synced up indexer and a synced up cardano-db-sync), then a questions:
- should it be a different test suite as it can be more naturally run separately that way?

Otherwise I guess using an env var to selectively run it, and passing a pattern to hedgehog to only run *that* test instead of the whole suite is the way.

(Also added the instruction document to reuse cardano-db-sync's node socket, that didn't make it to this repo. [This Jira issue](https://input-output.atlassian.net/browse/PLT-554))

Pre-submit checklist:
- Branch
    - [X] Tests are provided (if possible)
    - [X] Commit sequence broadly makes sense and have useful messages
- PR
    - [x] Self-reviewed the diff
    - [x] If relevant, reference the ADR in the PR and reference the PR in the ADR
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/marconi-chain-index.cabal
+++ b/marconi-chain-index/marconi-chain-index.cabal
@@ -22,7 +22,7 @@ source-repository head
 flag ci
   description: Toggle features on/off when in CI
   manual:      True
-  default:     False
+  default:     True
 
 common lang
   default-language:   Haskell2010
--- a/marconi-chain-index/marconi-chain-index.cabal
+++ b/marconi-chain-index/marconi-chain-index.cabal
@@ -19,6 +19,11 @@ source-repository head
   type:     git
   location: https://github.com/input-output-hk/marconi
 
+flag ci
+  description: Toggle features on/off when in CI
+  manual:      True
+  default:     False
+
 common lang
   default-language:   Haskell2010
   default-extensions:
@@ -342,7 +347,9 @@ test-suite marconi-chain-index-test-compare-cardano-db-sync
   type:           exitcode-stdio-1.0
   main-is:        Spec.hs
   hs-source-dirs: test-compare-cardano-db-sync
-  buildable:      False
+
+  if flag(ci)
+    buildable: False
 
   --------------------
   -- Local components
--- a/marconi-chain-index/marconi-chain-index.cabal
+++ b/marconi-chain-index/marconi-chain-index.cabal
@@ -262,7 +262,6 @@ benchmark marconi-chain-index-bench
 test-suite marconi-chain-index-test
   import:         lang
   ghc-options:    -Wno-unused-packages
-  ghc-options:    -fobject-code
   type:           exitcode-stdio-1.0
   main-is:        Spec.hs
   hs-source-dirs: test
--- a/marconi-chain-index/marconi-chain-index.cabal
+++ b/marconi-chain-index/marconi-chain-index.cabal
@@ -199,7 +199,6 @@ library marconi-chain-index-test-lib
     , plutus-core
     , plutus-ledger-api               ^>=1.1
     , plutus-tx
-    , plutus-tx-plugin
 
   ------------------------
   -- Non-IOG dependencies
@@ -261,7 +260,6 @@ benchmark marconi-chain-index-bench
 test-suite marconi-chain-index-test
   import:         lang
   ghc-options:    -Wno-unused-packages
-  ghc-options:    -fobject-code
   type:           exitcode-stdio-1.0
   main-is:        Spec.hs
   hs-source-dirs: test
@@ -307,7 +305,6 @@ test-suite marconi-chain-index-test
     , plutus-core
     , plutus-ledger-api               ^>=1.1
     , plutus-tx
-    , plutus-tx-plugin
 
   ------------------------
   -- Non-IOG dependencies
--- a/marconi-chain-index/marconi-chain-index.cabal
+++ b/marconi-chain-index/marconi-chain-index.cabal
@@ -265,7 +265,6 @@ test-suite marconi-chain-index-test
   main-is:        Spec.hs
   hs-source-dirs: test
   other-modules:
-    CompareToDbSync
     Spec.Marconi.ChainIndex.CLI
     Spec.Marconi.ChainIndex.Indexers.AddressDatum
     Spec.Marconi.ChainIndex.Indexers.AddressDatum.AddressDatumIndex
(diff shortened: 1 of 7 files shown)


berewt:
I don't know if we can easily exclude a test-suite from CI or not. Running them only if a variable is set sounds good enough for now.

berewt (approved):
LGTM.
Maybe we should document how to run the comparison test(s)?

(also, I prefer using `pure` over `return`, as it has less requirement, but it's a personal preference)

eyeinsky:
> LGTM. Maybe we should document how to run the comparison test(s)?

Yes, I guess I could start with writing it under the `doc/` folder?

eyeinsky:
Regarding having a separate cabal test suite component, [this is how it could be done from within .cabal file](https://input-output-rnd.slack.com/archives/CG1FBSDMM/p1681895211958379?thread_ts=1681895009.110199&cid=CG1FBSDMM):
```
Flag CI
  Description: Toggle features on/off when in CI
  Manual: True
  Default: False

...

if(CI)
  buildable: false
```

Why I was looking into this was that (IMO) it's more pleasant to run `cabal test marconi-chain-index-dbsync-compare` rather than `DBSYNC_COMPARE=1 cabal test marconi-chain-index-test`, and not need to handle the DBSYNC_COMPARE env variable from haskell code.

berewt:
I didn't know it was that easy to disable a test suite in the CI, it's indeed nicer.

eyeinsky:
I'm doing the epoch nonce ([PLT-1395](https://input-output.atlassian.net/browse/PLT-1395)) and epoch SDD ([PLT-387](https://input-output.atlassian.net/browse/PLT-387)) dbsync comparison issues together.

@koslambrou It looks like the epoch staking indexer is already correct, I guess the query to dbsync had previously some issue (perhaps related to deserialising pool ID from postgres or something)? The query to get stakepool sizes from dbsync is here: https://github.com/input-output-hk/marconi/pull/16/files#diff-e802208310c9d447d3cce1fb55c6800f55a3a5c453f0955c35bd9c23b453959aR73-R82

eyeinsky:
Another question which I just remembered: does Mamba have an example cardano-db-sync query for epoch sdd indexer (I couldn't find one in their scala code)?

marconi-chain-index/marconi-chain-index.cabal
```diff
@@ -259,10 +259,12 @@ benchmark marconi-chain-index-bench
 test-suite marconi-chain-index-test
   import:         lang
   ghc-options:    -Wno-unused-packages
+  ghc-options:    -fobject-code
```
koslambrou, line 262: Why was this needed?

And if you did need it, remove duplicate `ghc-options`?

marconi-chain-index/marconi-chain-index.cabal
```diff
@@ -215,6 +213,8 @@ library marconi-chain-index-test-lib
     , streaming
     , temporary
 
+  build-depends:   plutus-tx-plugin
```
koslambrou, line 216: Why move it here?

marconi-chain-index/test/CompareToDbSync.hs
```diff
@@ -0,0 +1,200 @@
...
+import Test.Tasty.Hedgehog (testPropertyNamed)
+
+
+tests :: TestTree
+tests = testGroup "Marconi to cardano-db-sync comparisons"
+  [ testPropertyNamed
+      "Compare *all* epoch nonces between Marconi and cardano-db-sync"
+      "propEpochNonce" propEpochNonce
+  , testPropertyNamed
+      "Compare random 10 epoch stakepool sizes between Marconi and cardano-db-sync"
```
koslambrou, line 50: Why 10 random? It's not better to compare **all** SDD for all epochs?

marconi-chain-index/test/CompareToDbSync.hs
```diff
@@ -0,0 +1,200 @@
...
+    \ GROUP BY epoch_no, pool_hash                  \
+    \ ORDER BY sum_amount desc                      "
+    $ PG.Only epochNo
+  return $ Map.fromList $ map (\(a, b) -> (a, rationalToLovelace b)) dbSyncRows
+  where
+    rationalToLovelace :: Rational -> C.Lovelace
+    rationalToLovelace n | 1 <- denominator n = fromIntegral $ numerator n
+                   | otherwise = error "getEpochStakepoolSizes: This should never happen, lovelace can't be fractional."
+
+queryIndexerStakepoolSiz :: C.EpochNo -> Storable.State EpochState.EpochStateHandle -> IO (Map.Map C.PoolId C.Lovelace)
```
koslambrou, line 90: queryIndexerStakepoolSiz -> queryIndexerStakepoolSize

koslambrou:
@eyeinsky Alright, did some manual tests. I think I used the wrong SQL in the past. Using the one from the Sidechains team yields the same results as marconi-sidechain :)

koslambrou:
For this PR, I think it's okay to put the tests in the module `ComparisonDbSync`. 

In the future, should we put the tests in `marconi-sidechain` instead and query using the JSON-RPC endpoints? Seems like this approach will test more components than by putting it in `marconi-chain-index`.

marconi-chain-index/marconi-chain-index.cabal
eyeinsky: Forgot to remove it -- it's required when one uses `cabal repl` but not for building.

marconi-chain-index/test/CompareToDbSync.hs
eyeinsky: All is better yes! It only takes a few minutes. Initially I thought it took much longer, but I was doing the "compare all" a hundred times (the hedgehog default number of tests)

eyeinsky:
I'm trying to make this a separate cabal component that is not executed by CI, asking advice here https://input-output-rnd.slack.com/archives/CG1FBSDMM/p1681895009110199.

@kayvank, you were setting up nix, maybe you have some knowledge, otherwise perhaps @zeme-iohk can help?

eyeinsky:
> In the future, should we put the tests in `marconi-sidechain` instead and query using the JSON-RPC endpoints? Seems like this approach will test more components than by putting it in `marconi-chain-index`.

In addition to that, it would be nice if both marconi and cardano-db-sync be accessible via HTTP (cardano-db-sync is only postgres) -- this way we could do the entire test over it and for cardano-db-sync we would not need to depend on postgres, and for marconi we wouldn't need setting up env variables for node socket, node config file and marconi db path (which are currently needed to create an indexer in haskell code).

koslambrou:
@eyeinsky Good idea. I remembered that there is the `cardano-graphql` library which is just that: a graphql interfact on top of cardano-db-sync. Probably something we want to integrate for easier setup. Then, moving the tests in `marconi-sidechain` will allow us to use the HTTP interface it provides. I'll create stories for those.

koslambrou (approved):

eyeinsky:
@koslambrou I am retarded -- there actually is a discrepancy with the Marconi indexer, my dbsync instance wasn't synced up to the top.. :facepalm: 

The difference starts from epoch 339, and although the number and IDs of the stakepools are the same for later epochs, then the amount of staked ADA isn't. Am continuing looking into this.

koslambrou:
@eyeinsky Which one is db-sync and which one is marconi?

PR #21, opened 2023-04-28
Add standalone index-state to cabal.project

zeme-iohk:
Seemingly redundant, but needed by certain versions/functions of haskell.nix, which would otherwise error out.


angerman:
It should not with recent Haskell.nix versions. Old ones did. New ones should not.

koslambrou:
@zeme-iohk angerman says  it's not needed for recent haskell.nix versions. Are we not recent enough?

zeme-iohk:
Yes, "you" are recent, but I'm trying to integrate the new CI here, and it failed. which means that _I_ am not recent enough. So we can close this

PR #24, opened 2023-05-03
Add ADR for the computation of the SDD per epoch

koslambrou:
Added an ADR describing the possible solutions for computing the SDD per epoch in `marconi-sidechain`.

The current implemented approach suffers from a large memory consumption because of the `LedgerState` computation which is used to extract the SDD.

We introduce different alternative solutions to the issue. All come with drawbacks. 

Hopefully, this will inform our decision on how to move forward.

Here is the rendered link of the document: https://github.com/input-output-hk/marconi/blob/PLT-5592-document-architectural-solutions-for-alternative-sdd-computation-for-marconi/marconi-sidechain/doc/adr/01-sdd-computation.adoc

### DIFF
--- a/marconi-sidechain/doc/adr/01-sdd-computation.adoc
+++ b/marconi-sidechain/doc/adr/01-sdd-computation.adoc
@@ -31,26 +31,24 @@ TBD
 
 We will describe all the potential and discussed solutions to index the SDD of each epoch in the following sections.
 
-As a summary, we provide the following caracteristics table and indicate how each solution responds to them.
-This assessment is measured in regards to two components: Cardano Node and Marconi.
+As a summary, we provide the following characteristics table and indicate how each solution responds to them.
+This assessment is measured in regard to two components: Cardano Node and Marconi.
 
-
-
-.Comparison of solutions using a list of caracteristics
+.Comparison of solutions using a list of characteristics
 [cols="1,1,1,1,1"]
 |===
-|Caracteristics | Marconi LedgerState computation | Read LedgerEvents from node | SDD using N2C | Marconi as a node
+|Characteristics | Marconi LedgerState computation | Read LedgerEvents from node | SDD using N2C | Marconi as a node
 
 |Number of LedgerState computations | 2 | 1 | 1 | 1
 |Initial node resync required | No | Yes | Yes | Yes
-|Required node resync after Marconi logic change | No | No | Yes | Yes
+|Required node resync after Marconi logic change | No | No | Yes | No
 |Change to cardano-node | No | Yes | No | No
 |===
 
-.Description of the caracteristics
+.Description of the characteristics
 [cols="1,1"]
 |===
-|Caracteristics | Description
+|Characteristics | Description
 
 |Number of LedgerState computations
 |As it suggest, corresponds to total number of independent computations of the LedgerState.
@@ -72,10 +70,6 @@ For more details, read the sections below.
 
 This is the approach that `cardano-db-sync` currently uses to compute the SDD for each epoch.
 
-A container diagram of this solution would look like:
-
-image::http://www.plantuml.com/plantuml/png/jL0_Q-Cm5D_zANIU4ZZ7SDnqKovkbz8w9g5pKAJNM4H_XFJKr9H-zqfAunfAe4ityNx_jVH87GObYrz2Cnaua1xH-dzrxUXfSH3OXttmu9ZH21eNpAZQIggJflB8fTZNpT-gIwV7X_jQKO_WucdNH0KL6jp2_eZ_3LXf1fL4EQ1e3OUPj-OtARVTcl0HB-VajIZ6FZEd8sK3aWuaQShv1DlR1AwaOKVIESC34_fmeQr4L6NvcHRBgg05emYUdEAMCXxRFosnpg1XHaRDn5lHD5D4piUhye83woCQYSbwPEIAJu2iZywPKVb1CmBFeD6VOvehCQDf-eZBc_otzXmonJyRCIfbzhtRhjjgqpNZt0dTNHdzjKtnLXJRmIEetUqBpY0h6-eurOPem9Dnnp4Xe-wx7xe4pOEIxm00[]
-
 Here is a general outline of the computation of the `LedgerState`:
 
 . At the start of `marconi-sidechain`, initialize the `LedgerState` (or `NewEpochState`) given the node genesis file
@@ -84,7 +78,11 @@ Here is a general outline of the computation of the `LedgerState`:
 . Compute the SDD from the new LedgerState and store in the database
 . Repeat from step 2
 
-While this approach is the "easiest" to implement, it suffers from major drawbacks such as:
+A container diagram of this solution would look like:
+
+image::http://www.plantuml.com/plantuml/png/jL0_Q-Cm5D_zANIU4ZZ7SDnqKovkbz8w9g5pKAJNM4H_XFJKr9H-zqfAunfAe4ityNx_jVH87GObYrz2Cnaua1xH-dzrxUXfSH3OXttmu9ZH21eNpAZQIggJflB8fTZNpT-gIwV7X_jQKO_WucdNH0KL6jp2_eZ_3LXf1fL4EQ1e3OUPj-OtARVTcl0HB-VajIZ6FZEd8sK3aWuaQShv1DlR1AwaOKVIESC34_fmeQr4L6NvcHRBgg05emYUdEAMCXxRFosnpg1XHaRDn5lHD5D4piUhye83woCQYSbwPEIAJu2iZywPKVb1CmBFeD6VOvehCQDf-eZBc_otzXmonJyRCIfbzhtRhjjgqpNZt0dTNHdzjKtnLXJRmIEetUqBpY0h6-eurOPem9Dnnp4Xe-wx7xe4pOEIxm00[]
+
+While this approach is the "easiest" to implement, it suffers from major drawbacks, such as:
 
 * adding an additional 10-16GB of RAM in `marconi-sidechain` just from computing the `LedgerState`
 * slowing down indexing.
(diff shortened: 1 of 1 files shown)


abailly-iohk (commented):
What could help clarify the decision is to have some kind of pictures representing the various solutions, along with some figures, eg. solution requires one node there and another node here, talking protocol X and requiring this amoiunt of RAM and Disk

marconi-sidechain/doc/adr/01-sdd-computation.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+The benefits:
+
+* no change to `cardano-node` are required
+* only one computation of `LedgerState` is required
+
+The drawbacks:
+
+* uncertainty of capability to use `cardano-node` as a library
+* this version of `marconi-sidechain` would use the same amount of memory as the relay node. Therefore, it would not be able to be deployed on a AWS small instance as initially requested by the Sidechain team
```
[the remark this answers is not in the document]
koslambrou: Good point, as we're currently focused on communicating with a local node. But I can imagine using Ogmios which could achieve that?

koslambrou:
> What could help clarify the decision is to have some kind of pictures representing the various solutions, along with some figures, eg. solution requires one node there and another node here, talking protocol X and requiring this amoiunt of RAM and Disk

Good idea!

marconi-sidechain/doc/adr/01-sdd-computation.adoc
```diff
@@ -0,0 +1,145 @@
...
+Here's list of assumptions:
+
+* `marconi-sidechain` will be run by a Stake Pool Operator (SPO)
+* `marconi-sidechain` will communicate with a relay node (not the block producing node)
+
+The main architectural goal we are trying to achieve is to reduce as much as possible the hardware resource usage of a SPO.
+
+Questions to be answered:
+
+* at what points in time (lower and upper bounds) does Sidechain need to know the SDD per epoch?
```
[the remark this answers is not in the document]
koslambrou: For now, it's `set`. Eventually, it might even be `mark`. However, that's probably out of scope for this ADR.

koslambrou:
> What could help clarify the decision is to have some kind of pictures representing the various solutions, along with some figures, eg. solution requires one node there and another node here, talking protocol X and requiring this amoiunt of RAM and Disk

@abailly-iohk I added a characteristic matrix for each solution, and a container diagram for each solution. Does that respond to your questions?

marconi-sidechain/doc/adr/01-sdd-computation.adoc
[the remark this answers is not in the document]
koslambrou: Ok, just confirmed with Sidechain team. They need the active SDD for any epoch.

berewt:
> > What could help clarify the decision is to have some kind of pictures representing the various solutions, along with some figures, eg. solution requires one node there and another node here, talking protocol X and requiring this amoiunt of RAM and Disk
> 
> @abailly-iohk I added a characteristic matrix for each solution, and a container diagram for each solution. Does that respond to your questions?

The table is an interesting summary
Can we provide a meaningful name to solutions instead of solution 1-4 (or a short summary of each solution before the table)? On first reads, it implies to go back and forth from the table to the solution description to get their description.

marconi-sidechain/doc/adr/01-sdd-computation.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+The benefits:
+
+* no change to `cardano-node` are required
+* uses the existing infrastructure to get the required information
+
+The drawbacks:
+
+* needs the SPO to resync the relay node from scratch
+* any indexing logic or database schema change to Marconi would require reindexing the local node (thus deleting the node database)
```
berewt, line 161: Just to clarify, is it because we won't be able to resume the SDD indexer?

marconi-sidechain/doc/adr/01-sdd-computation.adoc
```diff
@@ -0,0 +1,203 @@
...
+
+The benefits:
+
+* no change to `cardano-node` are required
+* only one computation of `LedgerState` is required
+
+The drawbacks:
+
+* uncertainty of capability to use `cardano-node` as a library
+* this version of `marconi-sidechain` would use the same amount of memory as the relay node.
```
berewt, line 182: But we won't need the other relay-node, so we move from one relay-node + marconi-sidechain to "just" Marconi-relay-node. Which probably even better (resource-wise) than relay-node + marconi-sidechain.

marconi-sidechain/doc/adr/01-sdd-computation.adoc
```diff
@@ -0,0 +1,203 @@
...
+As a summary, we provide the following caracteristics table and indicate how each solution responds to them.
+This assessment is measured in regards to two components: Cardano Node and Marconi.
+
+[cols="1,1,1,1,1"]
+|===
+|Caracteristics | Solution 1 | Solution 2 | Solution 3 | Solution 4
+
+|Number of LedgerState computations | 2 | 1 | 1 | 1
+|Initial node resync required | No | Yes | Yes | Yes
+|Required node resync after Marconi logic change | No | No | Yes | Yes
```
berewt, line 43: After reading the definition of solution 3 and 4 I don't get why we need a resync after logic change if it doesn't impact the SDD computation. Is it because we can't define, in the general case whether or not we will need to go through the SDD computation again to support the new logic?

eyeinsky:
Here's my (uninvited) feedback:

- Solution 3 is likely hackish to implement as it's hard to know when to do the SDD query at epoch boundaries. There might be a way to loop a query to the node to get the current epoch number, and then ask for the SDD when it changes.

  But still, the node would be running with full ledger state, and there is overhead to loop a "what is the epoch" query, and then ask for a SDD which is pushed through a serialize/deserialize cycle through the socket.

  Also, there is also the overhead of starting the indexer and the node at correct times, and also that there might still be a race condition when the node syncs ahead too fast, indexer is stalled for some reason, might miss an epochs, etc.

- Solution 2: I guess both `LedgerEvent`s and blocks could be considered as deltas that when applied to a previosu ledger state make up a new ledger state. Question is, will it be worth it to switch from applying blocks to applying `LedgerEvent`s.

  Specifically, with `LedgerEvent`s is it possible to keep a sub-ledger state, something that is smaller that the full ledger state to keep in memory? If not, then you'd still have the full ledger state in memory and only the thing that you apply is different (`LedgerEvent`s instead of blocks).

  Might still be worth it though if applying `LedgerEvent`s is somehow more efficient than applying blocks. Perhaps events are smaller, faster to apply, etc. On the other hand you'd have the overhead of creating those `LedgerEvent`s off-band -- blocks are already there.

- Solution 1: this is what we are trying to avoid if possible. Although I guess there would be value in having a good library for composing indexers -- to avoid "indexing everything". But having a second ledger state in RAM would still be an issue for many.

- Solution 4: seems like the best option. Whatever you do you'll always need a node that needs to sync up. When node and indexer are the same program, then you won't have any issues with timing queries nor sending stuff over the socket which are already present in the RAM of the same machine.

  cardano-node the executable should be well usable as a library (I looked at it, remember?). It's a wrapper around importing much of the code from consensus and ledger, and adds the logging part. Re-implementing a node inside Marconi though will be an undertaking regardless as the territory will be new.

(Don't know much about Mithrill, and can't comment on the last solution)

marconi-sidechain/doc/adr/01-sdd-computation.adoc
```diff
@@ -0,0 +1,203 @@
...
+. Get block from the chain sync protocol
+. Update the `LedgerState` given the block
+. Compute the SDD from the new LedgerState and store in the database
+. Repeat from step 2
+
+While this approach is the "easiest" to implement, it suffers from major drawbacks such as:
+
+* adding an additional 10-16GB of RAM in `marconi-sidechain` just from computing the `LedgerState`
+* slowing down indexing.
+  Computation of the `LedgerState` is _the_ bottleneck in `marconi-sidechain`
```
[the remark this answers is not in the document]
koslambrou: Yes, AFAIU, when selecting a committee for the next epoch, they need to query all the indexers such as EpochSDD, so that would not help. The _real_ problem is the memory footprint though, which I mentioned above.

marconi-sidechain/doc/adr/01-sdd-computation.adoc
koslambrou: Let's say, you start the relay node and marconi-sidechain from scratch. Imagine you get a change in requirements from the Sidechain tribe or you optimize the database schema in marconi-sidechain for faster syncing time and query response time. 

In those scenarios, you would need to reindex every in `marconi-sidechain` from scratch, right? Therefore, with this solution, reindexing everything from scratch would require the SPO to delete the node database, and rerun everything.

marconi-sidechain/doc/adr/01-sdd-computation.adoc
koslambrou: True :) Not really a drawback. I'll remove it.

marconi-sidechain/doc/adr/01-sdd-computation.adoc
koslambrou: If you change the indexing logic of `marconi-sidechain`, then you need to resync it from scratch, right? If that's the case, then you _also_ need to resync `cardano-node` from scratch for solutions 3 and 4.

koslambrou:
@eyeinsky Thanks for the review :) 

> Solution 3 ...

These are good points on why solution 3 is hacky. I'll add them to the ADR.

> Solution 2 ...Question is, will it be worth it to switch from applying blocks to applying LedgerEvents

Yes, I think so. If you take a look `LedgerEvent`, you'll notice that it gives you a `StakeDistEvent stakeMap` event, which is exactly what we need. The only issue is that the node doesn't store anywhere those `LedgerEvent`. Thus, that's why I noted that this solution would require a change in the `cardano-node` repo (or a fork).

If we were to go from applying blocks to SDD, then that is basically Solution 1. That's why we're proposing to use the `LedgerEvents` that could be emitted from the local node.

Did I misunderstand your point?

> Solution 4: seems like the best option. 

Yeah I think so. The only part I'm not sure is how much work would it be to access the `LedgerState` that's computed by the consensus layer in order to get the SDD.

marconi-sidechain/doc/adr/01-sdd-computation.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+* no change to `cardano-node` are required
+* uses the existing infrastructure to get the required information
+
+The drawbacks:
+
+* needs the SPO to resync the relay node from scratch
+* any change in the indexing logic of `marconi-sidechain` would require resyncing from scratch the local node (thus deleting the node database)
+* overhead of needing to continuously poll the relay node to identify when a new epoch occurs
+* race condition between node and `marconi-sidechain`. If a node syncs too fast and `marconi-sidechain` is stalled because of some heavy processing, we might miss an epoch (unlikely, but possible).
```
koslambrou, line 170: @eyeinsky I've updated this following your comment. Are your issues with Solution 3 sufficiently described here?

koslambrou: > I guess it would be hard to get the node (or the ledger) team to write ledger events to disk because it would be almost as if they did our work for us

Not exactly, because the data would not be queryable. Chain-indexers like Kupo or Marconi could take this data and index it in a db for fast querying.

> And if they did start to persist the events to disk, then soon enough we would show up with more requests, etc. They themselves don't care about the historic data to run the node, so for them the effort would all be for external bodies.

That's true. But there's nothing wrong with asking more to the node team. We are the customers and they are the product providers. How realistic it would be to actually get such a change accepted? I'm not sure myself. At the very least, we _could_ fork the node to provide such a functionality: which kind of goes into the idea of building custom nodes of Arnaud. That's why I think this solution could be as good as the solution for making Marconi as node. 

Don't forget, there is a major drawback of the solution of Marconi as a library: mainly that any change to the indexers requires a complete resync of the blocks and LedgerState.

marconi-sidechain/doc/adr/01-sdd-computation.adoc
berewt: We may find a way to resync only the parts that are impacted by a change of logic (especially with solution 4): so yes, we need to resync from scratch, but if we don't change the SDD computation we might keep our SDD indexing and have a lighter resync.
 It would be moving the reindexing problem we have in Marconi to a resync problem in a Marconi relay node, and thus limit the negative impact.

berewt: Oh but yes, in the case of side-chain, it would still impact the relay node, even if it's faster. I get the point now.

koslambrou:
@eyeinsky @berewt @kayvank Added a new possible solution with benefits and drawbacks: re-implement the ledger rule for keeping track of the SDD without relying on the `LedgerState`.

This solution brought a thought in my head. If, for good reason, we feel that this solution is not a suitable one, then that means that we are forced to use the `LedgerState` to compute the SDD. In that case, that also means that we should probably use the `LedgerState` to keep track of the UTXO set instead of re-implementing the UTXO ledger rule ourselves like we currently do. That would remove a lot of code and reduce the maintenance burden of the UTXO module.

Probably something to talk about in our tech discussion.

berewt (approved):
LGTM now, waiting for external feedbacks.

PR #28, opened 2023-05-05
Ml/plt 5474  better resume

eyeinsky:
This does both 
- [PLT-5474](https://input-output.atlassian.net/browse/PLT-5474): Return latest resumable chainpoint instead of entire database
- [PLT-5532](https://input-output.atlassian.net/browse/PLT-5532): Resume from oldest common resume point for indexers

Resuming happens as follows:
- every indexer reports the latest chain point it has indexed,
- the oldest of these is chosen as from where blocks are read over the socket
- indexers skip the blocks that they've already indexed ([`drainUntil`](https://github.com/input-output-hk/marconi/blob/a81a09b1672b2c343b0dba007e4473c281d260a3/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs#L525) is the helper function that does this)

Tests pass, all tests that were related to receiving a descending list of chain points from `resumeFromStorage` are removed.

---

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


eyeinsky:
(There is a commit that fixes the CI errors, once github is fully online then will push these, but the code can still be reviewed)

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -503,3 +521,22 @@ failWhenFull x = do
   if isEmpty
   then pure ()
   else error "an indexer raised an error"
+
+drainUntil :: C.ChainPoint -> TChan (ChainSyncEvent (BlockInMode CardanoMode)) -> QSemN -> IO ()
```
koslambrou, line 525: Add a comment here on what that function does and what it is used for.

Even better, add a {- [NOTE] ...} somewhere explaining the choice we made on handling resuming and link to it here.

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -503,3 +521,22 @@ failWhenFull x = do
...
+drainUntil cp tchan _barrier = go
+  where
+    go = do
+      signalQSemN _barrier 1
```
koslambrou, line 529: Why do we need this here?

koslambrou (approved):
Other than the minors comments, LGTM :)

marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/AddressDatum/AddressDatumIndex.hs
```diff
@@ -342,20 +320,11 @@ propResumingShouldReturnAtLeastOneNonGenesisPointIfStoredOnDisk = property $ do
...
-propResumablePointsShouldBeSortedInDescOrder = property $ do
-    cps <- forAll $ genChainPoints 1 5
-    events <- forAll $ forM cps genAddressDatumStorableEvent
-    indexer <-
-        liftIO $ raiseException $ AddressDatum.open ":memory:" (AddressDatumDepth 1)
-               >>= Storable.insertMany events
-               >>= Storable.insert (AddressDatum.toAddressDatumIndexEvent Nothing [] (last cps))
-    StorableProperties.propResumablePointsShouldBeSortedInDescOrder indexer
+    resumablePoint <- liftIO $ raiseException $ Storable.resumeFromStorage $ finalIndex ^. Storable.handle
+    let penultimate = case reverse $ List.sort cps of
```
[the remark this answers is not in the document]
eyeinsky: Added comment, 65c03abebd99b86f4498ef729965af19bf58f142

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
eyeinsky: b79ce37bf378295f1d651e9dd12876a74dbe93a8

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
eyeinsky: It's to signal the coordinator to send new blocks, it stalls otherwise.

berewt:
> * indexers skip the blocks that they've already indexed ([`drainUntil`](https://github.com/input-output-hk/marconi/blob/a81a09b1672b2c343b0dba007e4473c281d260a3/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs#L525) is the helper function that does this)

I think we discussed it but maybe it was while we were pair-programming with Kayvan. This makes the strong assumption that we never send to disk volatile event (otherwise we can skip the rollback while resuming).

I have mixed feeling about it: we may want to enforce it, but at the same time users may want to keep less events in memory and send volatile events to disk, accepting the risk of a disk-rollback, this won't be doable anymore.

We have 3 options:
- enforce that we don't store on-disk events that are non-volatile (the current assumption)
- rollback all indexers to the resume point instead of using `drainUntil`
- compute the last non-volatile block and rollback to this block indexers that are ahead and `drainUntil` we reach this block.

eyeinsky:
> I think we discussed it but maybe it was while we were pair-programming with Kayvan. This makes the strong assumption that we never send to disk volatile event (otherwise we can skip the rollback while resuming).

But a chain point is identified by a slot number and a block header hash, so if the hash matches it must be the same point for all to resume from. So indexers _can_ have volatile blocks on disk, and we only resume if it's the same chain point for all, no?

berewt:
> > I think we discussed it but maybe it was while we were pair-programming with Kayvan. This makes the strong assumption that we never send to disk volatile event (otherwise we can skip the rollback while resuming).
> 
> But a chain point is identified by a slot number and a block header hash, so if the hash matches it must be the same point for all to resume from. So indexers _can_ have volatile blocks on disk, and we only resume if it's the same chain point for all, no?

But we only check if the hash match for the resume point. As it's a minimum, indexers can be ahead of the resume point (it's why we have `drainUntil`) and we don't check if a rollback happen between the resume point and the current state fo the indexers.

eyeinsky:
If we resume from that minimal point then if there were a rollback after that, then I assume the node will send us that. Or do you mean that while we were not listening on the socket, there was a rollback but a one that didn't reach back as far as the resume point, and so we would not see it happening?

berewt:
> Or do you mean that while we were not listening on the socket, there was a rollback but a one that didn't reach back as far as the resume point (so the resume point itself is still valid), and so we would not see it happening? So indexers that were further ahead wouldn't observe that rollback.

Yes, this scenario.

eyeinsky:
I think I remember that whenever resuming from some chain point then the first chainsync event that comes in is always a rollback to that point. If this is so then the draining part of code won't be executed and your option 2 is what actually happens.

Hmm, then a way to go forward is to after getting the minimal resume point to also get all points from that to the latest point, and then zip those with what is received from the node, drain until they are equal (where the hash is also equal), stop draining when they diverge, send a rewind to the latest common point (such that everything forward is wiped) and continue from there.

eyeinsky:
I'm going to try find out whether it's part of the specification that the first chain sync event sent is always a rollback to the requested chain-point. If it is then it might be that we just pass on events to indexers with no filtering. (We could possibly throw an error when the first event isn't a rollback though)

eyeinsky:
Nevermind, we actually send the rollback ourselves:
https://github.com/input-output-hk/marconi/blob/72fa4c48f6e03956c2324283236e998a8ec659df/cardano-streaming/src/Cardano/Streaming.hs#L118

This is ready for review now.

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -120,44 +120,48 @@ initialCoordinator indexerCount' =
...
         RollBackward cp _ct -> do
           events <- Ix.getEvents (index ^. Ix.storage)
-          innerLoop ch $
+          innerLoop $
             fromMaybe index $ do
               slot   <- chainPointToSlotNo cp
               offset <- findIndex (any (\(s, _) -> s < slot)) events
               Ix.rewind offset index
 
+    resumeFromCp = C.ChainPointAtGenesis
```
berewt, line 146: I know you've just replicated the existing logic, but shouldn't we compute the resumePoint from the content of the indexer?

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -244,11 +249,12 @@ scriptTxWorker_
   -> Coordinator
   -> TChan (ChainSyncEvent (BlockInMode CardanoMode))
   -> FilePath
-  -> IO (IO (), MVar ScriptTx.ScriptTxIndexer)
+  -> IO (IO (), C.ChainPoint, MVar ScriptTx.ScriptTxIndexer)
```
berewt, line 252: Why did you bin the indexer from the return type in the Utxo indexer and not here? it looks like it's unused, so you may drop it here as well?

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -284,7 +290,7 @@ epochStateWorker_
   -> Coordinator
   -> TChan (ChainSyncEvent (BlockInMode CardanoMode))
   -> FilePath
-  -> IO (IO b, MVar EpochStateIndex)
+  -> IO (IO b, C.ChainPoint, MVar (Storable.State EpochStateHandle))
```
berewt, line 293: Same question here, the MVar may be binned.

berewt (approved):
LGTM, just a few cleanup to consider.

koslambrou:
The situation I'm concerned is: what if some indexers (similar to what EpochState does) can only resume from very specific slots. Normally, I would expect that if these resumable slots don't overlap, then we would be to restart from genesis. But what actually happens in the current implementation?

Did we decide on how to handle that resuming scenario?

berewt:
> The situation I'm concerned is: what if some indexers (similar to what EpochState does) can only resume from very specific slots. Normally, I would expect that if these resumable slots don't overlap, then we would be to restart from genesis. But what actually happens in the current implementation?
> 
> Did we decide on how to handle that resuming scenario?

We didn't and it's actually a big issue. I think none of the scenario we described address this constraints (except the original one, which doesn't scale).

berewt:
I've just checked how it's done with the experimental indexers: take the least advance chain point, rollback to it, and if the rollback fails, go back to genesis. It should work in this scenario. It would be by far better if we could check if a rollback is needed for each indexer (for EpochState for example, we never store non volatile event and thus we could just wait - using the now removed `drainUntil`, instead of trying to rollback, failing, and go back to genesis).

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
eyeinsky: The datum indexer seemed to be abandoned as it hasn't been updated to the newer Storable instance. Do you think I should write straight sql in here, or have custom helper in Datum's own module and use that?

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
eyeinsky: Hmm, what do you think about how to access the running indexer in general, would it be better to use the `onInsert` callback, or returning the indexer MVar? We haven't really decided on this and that is why the return types are a bit of a mess.

berewt: My understanding is that we use callback for this purpose (it's how it's done in Marconi-sidechain to expose the indexers to json-rpc queries iirc). In that case, we can drop it in the return type.
I would suggest to try to remove them and if it breaks anything, revert.

eyeinsky:
Just to summarize, what we try to improve in this PR was to avoid loading the entire database for to do a resume. The issue of if the epoch state indexer has a specific resume point isn't really resolved.

What happens when resuming multiple indexers where one of those is epoch state, is that if epoch state indexer doesn't have the oldest resume point then the entire thing will [error here](https://github.com/input-output-hk/marconi/blob/051d8895b309a599925d9a9a3cfe93a6e67fc6c9/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs#L358) (if the rollback point isn't found, but that is likely as there are comparatively few locations for which a ledger state exists on disk), so the process will just die.

The entire situation is also difficult to assess as we don't have any tests to run multiple indexers together. Hmm, what should we do here?

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
eyeinsky: It's the tests that rely on the indexer being returned, so now actually did need to revert the changes.

koslambrou:
> The entire situation is also difficult to assess as we don't have any tests to run multiple indexers together. Hmm, what should we do here?

Sorry, my comment was mainly about the general resuming problem. AFAIK, this PR was about finding a simple solution to fix the memory issue when resuming. I'm fine to merge the PR. Then, we would need to discuss the best way to do this in the future. That okay?

eyeinsky:
@koslambrou Rebased and will merge when CI passes (locally the test suite passed).

PR #35, opened 2023-05-16
Separate extract datum

eyeinsky:
[PLT-5568](https://input-output.atlassian.net/browse/PLT-5568): Move datum extraction logic of tx in a separate module

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


eyeinsky:
@koslambrou There were no tests that just used datum type/location without indexing it, so if a test module were to be added with tests that don't reference an indexer, then these tests would need to be created. Otherwise the test suite continues to pass (this change only moves code around, with no functional changes).

marconi-chain-index/src/Marconi/ChainIndex/Extract/Datum.hs
```diff
@@ -0,0 +1,41 @@
...
+
+import Cardano.Api qualified as C
+import Cardano.Api.Shelley qualified as C
+import Cardano.Ledger.Alonzo.TxWits qualified as Ledger
+import Data.Map.Strict (Map)
+import Data.Map.Strict qualified as Map
+import Data.Maybe (mapMaybe)
+
+-- | Get a map of datum hash to datum from a list of transactions.
+txsPlutusDatumsMap :: Foldable f => f (C.Tx era) -> Map (C.Hash C.ScriptData) C.ScriptData
```
koslambrou, line 14: I suggest to change to something like `getPlutusDatumsFromTxs` or `getPlutusResolvedDatumsFromTxs`.

koslambrou (commented):
Generally, I think we need to change the implementation. First off, we should have `getPlutusDatumsFromTxs` which returns resolved datums from a set of txs. Currently, the function only returns datums from the tx body. However, we want to include in there the inline datums as well.

For `getDatumsPerAddressFromTx`, I'm not sure about the signature. Maybe it should just return `Map AddressAny (Hash ScriptData)`? Then, we can have another function like `getResolvedDatumsFromTxs` which returns `Map AddressAny (Either (C.Hash C.ScriptData) (C.Hash C.ScriptData, C.ScriptData))`?

I'd like to have your thoughts on this idea.

koslambrou:
> There were no tests that just used datum type/location without indexing it, so if a test module were to be added with tests that don't reference an indexer, then these tests would need to be created. Otherwise the test suite continues to pass (this change only moves code around, with no functional changes).

In the module `Spec.Marconi.ChainIndex.Indexers.AddressDatum.AddressDatumIndexEvent`, there are generators that you can use (maybe need to adapt a bit) to write the property tests for those functions.

You can have a generator like:

```
genTxBodyWithDatums
    :: [DatumLocation]
    -> Gen (C.TxBody C.BabbageEra)
```

Then a property test like (pseudo-code):

```
   datumLocations <- forAll $ Gen.list (Range.linear 0 10) genDatumLocation
   txBody <- forAll $ genTxBodyWithDatums datumLocations
   txs <- forAll $ Gen.list (Range.linear 0 10) $ makeSignedTransaction txBody
   actualDatumMap <- getPlutusDatumsFromTxs txs
   let expectedDatumMap = undefined -- Convert the [DatumLocation] to `Map (Hash ScriptData) ScriptData`
   actualDatumMap === expectedDatumMap
```

What do you think?

angerman:
force-pushed to re-excite CI.

eyeinsky:
@koslambrou This ticket will help with the following issues:
- [PLT-5570](https://input-output.atlassian.net/browse/PLT-5570): Update Address-Utxo indexer to store resolved datums in a separate table
- [PLT-5569](https://input-output.atlassian.net/browse/PLT-5569): The Address-Utxo indexer query response should return a resolved datum

I would help if I could start perhaps doing those directly as it's otherwise difficult to refactor without a specific target. The current ticket talks about moving code around, and although I have a mild hunch on why, then doing a refactoring separately from code that will actually use this change, is hard.

eyeinsky:
@koslambrou Is it so that after all three Jira issues the AddressDatum indexer would be included in the Utxo indexer similar to how epoch nonce and sdd are a single indexer? Because if there is a query to the Utxo indexer where the result is required to have all datums resolved then this can be done by the help of the datum hash -> datum table, which has all past datums as well. If the Utxo and AddressDatum indexers are separate indexers (= have separate database files) then the join might be hard to do (and they can also be out of sync, although I guess only by one block by how the coordinator currently works).

I also see that [Utxo indexer does part of the datum extraction work](https://github.com/input-output-hk/marconi/blob/main/marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs#L1005-L1012) (which we can now import from the common module), but for some utxos there would only be the hash, and the datum itself might have been in a previous transaction.

eyeinsky:
Yup, got it!

eyeinsky:
@kayvank @berewt In the now two-table setup, is every row in `spent` table also matched by txId/txIx in `unspent_transactions` table? I'm wondering if the [`FULL OUTER JOIN` here is superfluous](https://github.com/input-output-hk/marconi/blob/main/marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs#L812) and a `LEFT JOIN` would do?

berewt:
> @kayvank @berewt In the now two-table setup, is every row in `spent` table also matched by txId/txIx in `unspent_transactions` table? I'm wondering if the [`FULL OUTER JOIN` here is superfluous](https://github.com/input-output-hk/marconi/blob/main/marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs#L812) and a `LEFT JOIN` would do?

Yes a left join should be enough, you can change it.

marconi-chain-index/src/Marconi/ChainIndex/Extract/Datum.hs
```diff
@@ -0,0 +1,105 @@
+{-# LANGUAGE DerivingStrategies #-}
```
koslambrou, line 1: I think this module should only contain datum extraction functions from transactions. The `DatumRow` and `createTable` should maybe not be here?

marconi-chain-index/src/Marconi/ChainIndex/Utils.hs
```diff
@@ -64,3 +66,7 @@ chainPointOrGenesis :: [C.ChainPoint] -> C.ChainPoint
 chainPointOrGenesis result = case result of
   []     -> C.ChainPointAtGenesis
   cp : _ -> cp
+
+-- | Convert a list of target addresses to a predicate function
+addressesToPredicate :: Maybe TargetAddresses -> Maybe (C.Address C.ShelleyAddr -> Bool)
```
koslambrou, line 71: That should maybe go in the Utxo indexer? That's the only place we're using it.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -215,93 +225,81 @@ instance ToJSON Utxo where
             , "txIndexInBlock"    .= (u ^. txIndexInBlock)
             ]
 
-data ChainPointRow
-    = ChainPointRow { _cpSlotNo :: C.SlotNo, _cpBlockHash :: C.Hash C.BlockHeader }
-    deriving (Show, Eq, Ord, Generic)
```
koslambrou, line 220: Why did you delete it?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -215,93 +225,81 @@ instance ToJSON Utxo where
...
     } deriving (Show, Eq, Ord, Generic)
 
 $(makeLenses ''SpentInfo)
 
 data UtxoRow = UtxoRow
-  { _urUtxo          :: !Utxo
-  , _urCreationPoint :: !ChainPointRow
-  , _urSpentInfo     :: !(Maybe SpentInfo)
+  { _urUtxo            :: !Utxo
+  , _urSlotNo          :: !C.SlotNo
```
koslambrou, line 239: I like the old names: `urCreationSlotNo` and `urCreationBlockHeaderHash`.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -215,93 +225,81 @@ instance ToJSON Utxo where
...
+  { utxoResultAddress          :: !C.AddressAny
+  , utxoResultTxIn             :: !C.TxIn
+  , utxoResultDatum            :: !(Maybe C.ScriptData) -- datumhash_datum
+  , utxoResultDatumHash        :: !(Maybe (C.Hash C.ScriptData))
+  , utxoResultValue            :: !C.Value
+  , utxoResultInlineScript     :: !(Maybe C.ScriptInAnyLang)
+  , utxoResultInlineScriptHash :: !(Maybe C.ScriptHash)
+  , utxoResultTxIndexInBlock   :: !TxIndexInBlock
+  --
+  , utxoResultSlotNo           :: !C.SlotNo
```
koslambrou, line 256: `utxoResultSlotNo` -> `utxoResultCreationSlotNo` as above?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -215,93 +225,81 @@ instance ToJSON Utxo where
...
+  , _urSlotNo          :: !C.SlotNo
+  , _urBlockHeaderHash :: !(C.Hash C.BlockHeader)
+  , _urSpentInfo       :: !(Maybe SpentInfo)
   } deriving (Show, Eq, Ord, Generic)
 
 $(makeLenses ''UtxoRow)
 
-urSpentSlotNo :: Traversal' UtxoRow C.SlotNo
-urSpentSlotNo = urSpentInfo . _Just . siSpentPoint . cpSlotNo
+data UtxoResult = UtxoResult_
```
koslambrou, line 246: UtxoResult_? Why not just `UtxoResult`?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -215,93 +225,81 @@ instance ToJSON Utxo where
...
+    , "slotNo" .= _urSlotNo ur
+    , "blockHeaderHash" .= _urBlockHeaderHash ur
     , "spentSlotNo" .= preview urSpentSlotNo ur
     , "spentBlockHeaderHash" .= preview urSpentBlockHash ur
     , "spentTxId" .= preview urSpentTxId ur
     ]
 
 data instance StorableResult UtxoHandle
-    = UtxoResult { getUtxoResult :: ![UtxoRow] }
+    = UtxoResult { getUtxoResult :: [UtxoResult] }
```
koslambrou, line 302: Use strict? `![UtxoResult]`

koslambrou (commented):
I don't know if it's me, but there seems to be a lot of changes that seem to me out of scope for the story in question. Seems like a lot of refactoring going on. Is that needed for completing the story?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
eyeinsky: This seemed to be an internal data type that added boilerplate, while there is no table for the chain point rows. I.e there didn't seem to be a reason to re-wrap the chain point, so I inlined it into the two places it was used.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
eyeinsky: 9e799af49d9b82b99ced8dde8fcd0e0c46b4eb4f

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
eyeinsky: The `UtxoResult` was taken by the query.. Maybe another name to avoid the underscore entirely?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
eyeinsky: 79de6fc6c6e69838b462e959c2117e1ae55e4e79

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
eyeinsky: fb05023dc (the rest of the list will still be lazy though)

marconi-chain-index/src/Marconi/ChainIndex/Utils.hs
eyeinsky: It's was used in the AddressDatum indexer, the move out was to re-use it in the Utxo indexer. (Maybe all the refactor looks better after it's closer to complete, as now it was half way there and looks funny as it's only used in one single other place.)

marconi-chain-index/src/Marconi/ChainIndex/Extract/Datum.hs
eyeinsky: Right, so maybe it doesn't look like it belongs to an extract module, but as the table has the same function between Utxo indexer and AddressDatum then then we could reuse it, no? (Maybe call the module Marconi.ChainIndex.Datum?)

eyeinsky:
@koslambrou Pushed where I got to:

 - Added the datum hash/datum map to Utxo event.
- Added the `UtxoResult` which is used only to _read_ from the sqlite database and reply to the query. (The UtxoRow and DatumRow are used to write to the database). The data type is separate because sql can already do the join querying the data (so we don't need to loop an sql query for each UtxoRow to fetch the datum)
- Regarding refactoring, some of it was just removing things while studying code that (IMO) didn't add much: e.g the ChainPointRow is just a chain point which can be within data types directly and then doesn't require sql, json and lens instances.

There are a few TODOs left (which I didn't have time for :():
- one failing test: `Roundtrip events to sqlRows transformation property`
- UtxoResult needs a ToJSON instance for the marconi-chain-index-bench (`cabal build all` reveals the error)
- and perhaps a test for an Utxo event with added datums that are resolved (to have tests which more tests which fail with datums not indexed)

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
koslambrou: Also, why define `UtxoResult` vs use the already existin `UtxoRow`?

eyeinsky: After the rebase I've forgotten to remove the datum field from `Utxo` so that is why `UtxoRow` and `UtxoResult` are still isomorphic.. But the idea is for talking to sqlite to have different data types for incoming and outgoing data: when writing to the database then there are two[1] tables: `unspent_transactions` and `datumhash_datum`, which both should have their own row types (`UtxoRow` and `DatumRow`). And then there is the sql select that is used to reply to indexer queries which has all the tables joined, this is the `UtxoResult`.

[1] Actually, `spent` is now back too, but adding this could be another jira ticket.

eyeinsky:
@koslambrou @berewt I have a question: since if we store datums in a separate table from tx outputs in the `unspent_transactions` then effectively the `getStoredEvents` function won't work anymore (?) as we've lost the information on whether the datum in the txOut was inline or not, as they are now all stored by hash in the `unspent_transactions` table. In other words writing to sqlite with `persistToStorage` and `getStoredEvents` won't match anymore (unless we write into the `unspent_transactions` table also _how_ the datum was in the txOut).

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
koslambrou: Ahh I understand now. In that case, yes, your solution makes sense.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
koslambrou: :+1:

eyeinsky:
@kayvank Could you help me understand [this test](https://github.com/input-output-hk/marconi/blob/b80515cba2fb751eaca1992bebc37f14c5d33647/marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs#L123-L183). It says in the description that all queried utxos should be unspent (if I understand correctly). But then when I look at how the mockchain is generated then that seems to pick a [random subset](https://github.com/input-output-hk/marconi/blob/b80515cba2fb751eaca1992bebc37f14c5d33647/marconi-chain-index/test-lib/Gen/Marconi/ChainIndex/Mockchain.hs#L58) of all previously seen utxos as inputs for the next transaction. This means (IMO) that new transactions spend earlier utxos.

I haven't been able to figure out: what guarantees that in this particular test none of the utxos will be spent?

kayvank:
ns

> @kayvank Could you help me understand [this test](https://github.com/input-output-hk/marconi/blob/b80515cba2fb751eaca1992bebc37f14c5d33647/marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs#L123-L183). It says in the description that all queried utxos should be unspent (if I understand correctly). But then when I look at how the mockchain is generated then that seems to pick a [random subset](https://github.com/input-output-hk/marconi/blob/b80515cba2fb751eaca1992bebc37f14c5d33647/marconi-chain-index/test-lib/Gen/Marconi/ChainIndex/Mockchain.hs#L58) of all previously seen utxos as inputs for the next transaction. This means (IMO) that new transactions spend earlier utxos.
> 
> I haven't been able to figure out: what guarantees that in this particular test none of the utxos will be spent?



> @kayvank Could you help me understand [this test](https://github.com/input-output-hk/marconi/blob/b80515cba2fb751eaca1992bebc37f14c5d33647/marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs#L123-L183). It says in the description that all queried utxos should be unspent (if I understand correctly). But then when I look at how the mockchain is generated then that seems to pick a [random subset](https://github.com/input-output-hk/marconi/blob/b80515cba2fb751eaca1992bebc37f14c5d33647/marconi-chain-index/test-lib/Gen/Marconi/ChainIndex/Mockchain.hs#L58) of all previously seen utxos as inputs for the next transaction. This means (IMO) that new transactions spend earlier utxos.
> 
> I haven't been able to figure out: what guarantees that in this particular test none of the utxos will be spent?

- The generator, as you mentioned, guarantees there will be at least one Utxo, possibly more.  
- The test claims, if we insert all generated events, and query for all of them, there will be at least one legitimate Utxo, meaning it is truly Utxo and therefor has no 'Spent'.
- The garantee comes from merging the in-memory & storage performed by the [mergeInMemoryAndSql function](https://github.com/input-output-hk/marconi/blob/b80515cba2fb751eaca1992bebc37f14c5d33647/marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs#L768)

Hope that helped?

eyeinsky:
@koslambrou I think I found and fixed a bug (resolve datums for memory events against the database) but it didn't make the test succeed, hmm :thinking:

eyeinsky:
The following succeeds, so it's the "plutus" location:
```haskell
  let datGen =
        Gen.choice
          [ pure Gen.NoDatumLocation
          , fmap (\d -> Gen.TxOutDatumHashLocation (C.hashScriptDataBytes d)) CGen.genHashableScriptData
          , fmap (\d -> Gen.TxOutDatumInTxLocation (C.hashScriptDataBytes d) d) CGen.genHashableScriptData
          , fmap (\d -> Gen.TxOutDatumInlineLocation (C.hashScriptDataBytes d) d) CGen.genHashableScriptData
          -- , fmap (\d -> Gen.PlutusScriptDatumLocation (C.hashScriptDataBytes d) d) CGen.genHashableScriptData
          ]
```

koslambrou:
> @koslambrou I think I found and fixed a bug (resolve datums for memory events against the database) but it didn't make the test succeed, hmm thinking

Yeah I noticed it. Also, there was missing the Datum.insertRows statement in persistToStorage. 

I updated the test case. Still failing, but this time the test looks correct. Basically, the failing scenario is when tx1 has an script output with a datum hash. Then tx2 in a separate block tries to spend that output by providing the datum in the Plutus datum witness. The test says the UtxoResult doesn't contain the expected datum.

Edit:

Mmm... maybe that's normal. If the utxo is spent, it shouldn't be returned by the query result. I guess what that means is that the result will never contain a resolved datum that was provided as a datum witness when spending the output.
I'll fix the test case tomorrow.

eyeinsky:
> Mmm... maybe the test is wrong. If the utxo is spent, it shouldn't be returned by the query result. I guess what that means is that the utxo result will never contain a resolved datum for a datum that was resolved by spending the UTXO. I'll fix the test case tomorrow.

We could ask Mamba, perhaps during a meeting as the issue is intricate. In the query there is a slot defined, so do they want utxos with datums resolved by that point, or is it fine to resolve all datums, even if they were provided in the future (future = later than the slot provided in the query).

Regarding the missing Datum.insertRows, would it be ok do comment out the failing test for now and merge the current PR to main? As otherwise I make mistakes when resolving merge conflicts whenever merging main into the current PR and trying to keep up to date.

marconi-chain-index/test/Integration.hs
```diff
@@ -1,261 +0,0 @@
-{-# LANGUAGE DataKinds #-}
```
koslambrou, line 1: Why was this deleted?

eyeinsky: It's already in the module Spec.Marconi.ChainIndex.Indexers.ScriptTx, I think it was just forgotten to be deleted.

koslambrou (approved):

eyeinsky:
@berewt I enabled auto-merge as it's pending CI..

PR #37, opened 2023-05-18
PLT-41 Fixed resuming error in EpochState indexer when the LedgerState was partially written on disk.

koslambrou:
* In the resumablePoints implementation, we now try to deserialize the LedgerState when identifying the possible resumable points from the LedgerStates saved on disk. If the file is not deserializable, we ignore it. Then, that LedgerState will get deleted in the `persistToStorage` implementation.

* Added a proper datatype (`LedgerStateFileMetadata`) for representing the metadata contained in the LedgerState filename.

* Tested on mainnet and preprod. Validated that when stopping `marconi-chain-index` when LedgerState is only partially written, it deletes it and resumes from the previous one.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
@@ -55,6 +55,7 @@
 --   'LedgerState' if it's rollbackable or if the last one of a given epoch. This step is necessary
 --   for resuming the indexer.
 --   4. we delete immutable 'LedgerState' binary files expect latest one (this step is necessary for
+--   keeping the disk usage as low as possible).
 --
 -- The indexer provides the following queries:
 --
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
@@ -112,7 +112,7 @@ import Marconi.ChainIndex.Error (IndexerError (CantInsertEvent, CantQueryIndexer
                                  liftSQLError)
 import Marconi.ChainIndex.Orphans ()
 import Marconi.ChainIndex.Types (SecurityParam)
-import Marconi.ChainIndex.Utils (chainPointOrGenesis, isBlockRollbackable)
+import Marconi.ChainIndex.Utils (chainPointOrGenesis, getBlockNoFromChainTip, isBlockRollbackable)
 import Marconi.Core.Storable (Buffered (persistToStorage), HasPoint (getPoint), Queryable (queryStorage), Resumable,
                               Rewindable (rewindStorage), State, StorableEvent, StorableMonad, StorablePoint,
                               StorableQuery, StorableResult, emptyState)
@@ -181,7 +181,9 @@ toStorableEvent
     -> Bool -- ^ Is the last event of the current epoch
     -> StorableEvent EpochStateHandle
 toStorableEvent extLedgerState slotNo bhh bn chainTip securityParam isFirstEventOfEpoch = do
-    let doesStoreLedgerState = isBlockRollbackable securityParam bn chainTip || isFirstEventOfEpoch
+    let doesStoreLedgerState =
+            isBlockRollbackable securityParam bn (getBlockNoFromChainTip chainTip)
+            || isFirstEventOfEpoch
     EpochStateEvent
         (if doesStoreLedgerState then Just extLedgerState else Nothing)
         (getEpochNo extLedgerState)
@@ -344,11 +346,17 @@ instance ToJSON EpochNonceRow where
         ]
 
 data LedgerStateFileMetadata = LedgerStateFileMetadata
-    { lsfMetaIsVolatile      :: !Bool
-    , lsfMetaSlotNo          :: !C.SlotNo
+    { lsfMetaSlotNo          :: !C.SlotNo
     , lsfMetaBlockHeaderHash :: !(C.Hash C.BlockHeader)
     , lsfMetaBlockNo         :: !C.BlockNo
+    , lsfMetaChainTipBlockNo :: !C.BlockNo
     }
+
+isLedgerStateFileRollbackable :: SecurityParam -> LedgerStateFileMetadata -> Bool
+isLedgerStateFileRollbackable
+        securityParam
+        LedgerStateFileMetadata { lsfMetaBlockNo, lsfMetaChainTipBlockNo } =
+    isBlockRollbackable securityParam lsfMetaBlockNo lsfMetaChainTipBlockNo
 
 instance Buffered EpochStateHandle where
     -- We should only store on disk SDD from the last slot of each epoch.
@@ -389,15 +397,16 @@ instance Buffered EpochStateHandle where
         -- We store the LedgerState if one of following conditions hold:
         --   * the LedgerState cannot be rollbacked and is the last of an epoch
         --   * the LedgerState can be rollbacked
-        let writeLedgerState ledgerState (C.SlotNo slotNo) blockHeaderHash (C.BlockNo blockNo) isRollbackable = do
+        let writeLedgerState ledgerState (C.SlotNo slotNo) blockHeaderHash (C.BlockNo blockNo) (C.BlockNo nodeTipBlockNo) = do
                 let fname = ledgerStateDirPath
                         </> "ledgerState_"
-                         <> (if isRollbackable then "volatile_" else "")
                          <> show slotNo
                          <> "_"
                          <> Text.unpack (C.serialiseToRawBytesHexText blockHeaderHash)
                          <> "_"
                          <> show blockNo
+                         <> "_"
+                         <> show nodeTipBlockNo
                          <> ".bin"
                 -- TODO We should delete the file is the write operation was interrumpted by the
                 -- user. Tried using something like `onException`, but it doesn't run the cleanup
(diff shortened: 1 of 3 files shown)


marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
```diff
@@ -435,20 +444,22 @@ instance Buffered EpochStateHandle where
...
-                  mapMaybe (\fp -> fmap (fp,) $ chainTipsFromLedgerStateFilePath fp)
+                  mapMaybe (\fp -> fmap (fp,) $ readLedgerStateFileMetadata fp)
                   <$> listDirectory ledgerStateDirPath
 
               -- Delete volatile LedgerState which have become immutable.
               let oldVolatileLedgerStateFilePaths =
                       fmap fst
-                      $ filter (\(_, (isVolatile, _, _, blockNo)) ->
-                          isVolatile && not (isBlockRollbackable securityParam blockNo chainTip))
+                      $ filter (\(_, LedgerStateFileMetadata { lsfMetaIsVolatile, lsfMetaBlockNo }) ->
```
eyeinsky, line 453: In this entire delete section, if the goal is to keep just the latest non-volatile ledger state, then whether a ledger state was volatile or not at the time it was written is now outdated and can be ignored (?) -- just filter with `isBlockRollbackable` to remove the ledger states which are volatile _now_, and from the remaining list remove all the older ledger states.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
```diff
@@ -591,55 +591,69 @@ instance Rewindable EpochStateHandle where
...
+        liftSQLError CantQueryIndexer $ do
+            -- We first identify the chain points from the LedgerState files that we stored on disk.
+            -- We ignore volatile ledger state files and we try to deserialise the LedgerState to
+            -- ensure that we correctly serialised it in a previous run. If the file is not
+            -- deserializable, we delete it.
+            ledgerStateFilepaths <- listDirectory ledgerStateDirPath
+            readableLedgerStateFilePaths <- flip filterM ledgerStateFilepaths $ \ledgerStateFilePath -> do
+                lsM <- readLedgerStateFromDisk (ledgerStateDirPath </> ledgerStateFilePath) topLevelConfig
+                case lsM of
+                  Nothing -> pure False
```
eyeinsky, line 615: Maybe delete the file already here? It won't be any use later on anyway..

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
```diff
@@ -42,22 +42,24 @@
...
 --
 --   1. we save the SDD for the current epoch in the `epoch_sdd` table
---   2. we save the 'LedgerState's in the filesystem as binary files (the ledger state file path has
+--   2. we save the Nonce for the current epoch in the `epoch_nonce` table
+--   3. we save the 'LedgerState's in the filesystem as binary files (the ledger state file path has
 --   the format: `ledgerState_<SLOT_NO>_<BLOCK_HEADER_HASH>_<BLOCK_NO>.bin`). We only store a
 --   'LedgerState' if it's rollbackable or if the last one of a given epoch. This step is necessary
 --   for resuming the indexer.
---   3. we delete immutable 'LedgerState' binary files expect latest one (this step is necessary for
+--   4. we delete immutable 'LedgerState' binary files expect latest one (this step is necessary for
```
eyeinsky, line 57: unfinished parenthesis

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
```diff
@@ -340,6 +342,13 @@ instance ToJSON EpochNonceRow where
         , "blockNo" .= blockNo
         ]
 
+data LedgerStateFileMetadata = LedgerStateFileMetadata
+    { lsfMetaIsVolatile      :: !Bool
```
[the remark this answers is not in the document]
koslambrou: Actually, this is determined by the filename of the LedgerState (as it has the `volatile` word in it). Are you suggesting that we instead add the current node tip in the filename of the LedgerState, and then automatically do the computation to determine if the LedgerState is volatile or not?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
koslambrou: > if the goal is to keep just the latest non-volatile ledger state

No, the goal is to keep the latest non-volatile LedgerState *and* all volatile LedgerStates.

> just filter with isBlockRollbackable to remove the ledger states which are volatile now,

But I don't want to filter LedgerStates which are volatile now, I need to keep them in case of a rollback.

However, your point does bring up that something is maybe wrong here... 
In the first deletion part, I'm removing all volatile LedgerStates that have become immutable. However, that's seems wrong AFAIU. I need to *rename* the file from `ledgerState_volatile_...` to `ledgerState_...` and then the next delete part will delete all LedgerStates except the latest one.

Does that make this more valid?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
koslambrou: The deletion logic is already implemented in `rewindStorage`. Once we start the chain-sync protocol with the point we're going to resume from, a `RollBackward` event is emitted, which in our case will delete the corrupt LedgerState given the implement of `rewindStorage`. I thought that I wouldn't want to duplicate the logic here. Does that make sense to you?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
koslambrou: Thanks. See here aea42008f81a00feda2afa0469c77ec9e8a23d07

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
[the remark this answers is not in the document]
eyeinsky: I was suggesting to leave the volatile info out because the fact of whether a ledger state is volatile or not changes in time (when current tip is more than security parameter blocks away from the slot of the ledger state). So ledger states that once were volatile become non-volatile as the tip advances, maybe there is some work to be saved by not deleting them.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
eyeinsky: Yup, this now makes it clearer! 

What I was trying to get at at the other comment is to not add the `_volatile` tag at all, because whether the ledger state is volatile or not depends on where the current tip is when the program is using the file (the _now_ for the program), and not where the current tip was when the file was written. I.e maybe on of the volatile ledger states that now has become immutable (and is deleted) is actually the one that got written just a few lines above. So it's possible to save a write/delete cycle.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
eyeinsky: Isn't there a danger that whenever ledger states get removed then accidentally a corrupt one gets kept? Maybe the system as a whole functions in a way that this doesn't occur, but it seems like if nothing bad can happen when the file is deleted right away then just delete it right away.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
koslambrou: Alright that makes sense. I'll do the following changes:

* remove the "volatile" keyboard in the filename
* modify the deletion logic to only keep the latest immutable LedgerStates along with *all* volatile LedgerStates, and assess if a LedgerState is immutable using the available information

koslambrou: @eyeinsky I applied the suggestions here 479869bf8868db054f4c794b413eea492465fbaa
Can you validate?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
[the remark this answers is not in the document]
koslambrou: See 479869bf8868db054f4c794b413eea492465fbaa

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
koslambrou: See 479869bf8868db054f4c794b413eea492465fbaa

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
eyeinsky: Still can't the volatility of a ledger state be determined dynamically, from the "live" chain tip be gotten from the latest block that came in over chainsync instead? But I see now that in [resumeFromStorage](https://github.com/input-output-hk/marconi/pull/37/files#diff-50d35d89163b61b8994cdf3b2e505d43ec2f833875c43469c9c075053ebfdfa1R609-R654) the actual tip at the time of resuming isn't passed in, so with the current indexer API I guess this is the best we can do.

eyeinsky (approved):

marconi-chain-index/src/Marconi/ChainIndex/Indexers/EpochState.hs
koslambrou: That's right, it's unfortunately not possible to have it in `resumeFromStorage`. That is, it is possible to query the tip to the local node *before* running the indexers, and then pass that info in `EpochSPDHandle`. But then, we would have an information in `EpochSPDHandle` that will *only* be used for resuming and will be obsolete afterwards. Not sure if that's worth it. Thoughts?

PR #40, opened 2023-05-19
Add burn query to MintBurn indexer, query only burn in marconi-sidechain

eyeinsky:
[PLT-5577](https://input-output.atlassian.net/browse/PLT-5577): MintBurn indexer: get only transactions that resulted in burning of tokens:
- Added queries to get only burn events
- Added test to query burns, both all of them and then pick a random one and query that
- Use the burn query in marconi-sidechains JSON-RPC API
- Added sample query and response to README (this appears to be the first burn on mainnet)
- Changed tests to generate both mint and burn (quantity < 0) events, all tests continue to pass (except endToEnd, there I suspect one can't burn more than has been minted, so generated events would have to take that into account) 

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
@@ -293,6 +293,12 @@ data instance RI.StorableQuery MintBurnHandle
   | QueryAllMintBurn (Maybe C.SlotNo)
   -- ^ Query all transactions that minted 'AssetId's until an upper bound slot in the blockchain. If
   -- the upper bound slot is 'Nothing', then we return everything.
+  | QueryBurnByAssetId C.PolicyId C.AssetName (Maybe C.SlotNo)
+  -- ^ Query all transactions that burned a specific 'AssetId' until an upper bound slot in the
+  -- blockchain. If the upper bound slot is 'Nothing', then we return everything.
+  | QueryAllBurn (Maybe C.SlotNo)
+  -- ^ Query all transactions that burned 'AssetId's until an upper bound slot in the blockchain. If
+  -- the upper bound slot is 'Nothing', then we return everything.
   deriving (Show)
 
 newtype instance RI.StorableResult MintBurnHandle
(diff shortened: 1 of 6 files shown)


marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
```diff
@@ -306,56 +306,97 @@ data instance RI.StorableQuery MintBurnHandle
   | QueryAllMintBurn (Maybe C.SlotNo)
   -- ^ Query all transactions that minted 'AssetId's until an upper bound slot in the blockchain. If
   -- the upper bound slot is 'Nothing', then we return everything.
+  | QueryBurnByAssetId C.PolicyId C.AssetName (Maybe C.SlotNo)
+  -- ^ Query all burn transactions that minted a specific 'AssetId' until an upper bound slot in the
```
koslambrou, line 310: I don't think the comment is correct. Should be something like "Query all transactions that burned a specific 'AssetId' until an upper bound slot in the blockchain". Also, maybe we should say "Query all transaction events", instead of just transactions.

marconi-sidechain/README.md
```diff
@@ -195,7 +195,23 @@ $ curl -d '{"jsonrpc": "2.0" , "method": "getUtxosFromAddress" , "params": { "ad
...
+$ curl -d '{"jsonrpc": "2.0", "method": "getTxsBurningAssetId", "params": {"policyId": "fda1b6b487bee2e7f64ecf24d24b1224342484c0195ee1b7b943db50", "assetName": "4c6f6273746572436f756e746572"}, "id": 1}' -H 'Content-Type: application/json' -X POST http://localhost:3000/json-rpc | jq
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": [
+    {
+      "blockHeaderHash": "834ac116737ed735805864491626aa8eadca3a9c3be86b559c65c92fd3ddd9fa",
+      "quantity": -22411,
+      "redeemer": "80",
+      "redeemerHash": null,
```
koslambrou, line 208: Should not be null.

eyeinsky: Hmm, looks like we are not indexing it at all, it's set to Nothing [for all results](https://github.com/input-output-hk/marconi/blob/7cdedd3a4ec05f989b9d001a3676e57924376822/marconi-sidechain/src/Marconi/Sidechain/Api/Query/Indexers/MintBurn.hs#L75).

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
eyeinsky: Tried a better comment: 5e2f5911913a99e96c4994eed3297298da8e5074

marconi-sidechain/README.md
koslambrou: Yeah, we should compute it. Can you add it? Should not be complicated, it's just a call to `hashScriptData` from cardano-api.

eyeinsky: Will do!

eyeinsky: Added the hash 497179774, ran the query again to get the hash for the sample f4a652c66.

berewt (approved):
Just a few cosmetic remarks, it's up to you to include them or not.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
koslambrou: @eyeinsky  Needs addressing.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
```diff
@@ -293,56 +293,90 @@ data instance RI.StorableQuery MintBurnHandle
...
+          <> (["quantity < 0"], [])
+
+      mkAssetIdCondition :: C.PolicyId -> C.AssetName -> ([SQL.Query], [NamedParam])
+      mkAssetIdCondition policyId assetName =
+        ( ["policyId = :policyId", "assetName = :assetName"]
+        , [":policyId" := policyId, ":assetName" := assetName]
+        )
+
+      mkUpperBoundCondition :: Maybe C.SlotNo -> ([SQL.Query], [NamedParam])
+      mkUpperBoundCondition maybeUpperBound = case maybeUpperBound of
```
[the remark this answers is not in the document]
eyeinsky: How do you mean? (It's already a case statement)

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
```diff
@@ -293,56 +293,90 @@ data instance RI.StorableQuery MintBurnHandle
...
+        Nothing -> ([], [])
```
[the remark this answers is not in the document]
eyeinsky: Since the function already returns a monomorphic list then this would indirectly still mean `[]`.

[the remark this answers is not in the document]
berewt: oh no i meant instead of the whole `([],[])` stuff. And yes, `mempty` would mean the exact same thing but as we consider to append it in the end using `<>`, `mempty` in this specific case may be clearer (but I know that we may have different opinions on this so it's up to you to decide).

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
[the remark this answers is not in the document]
berewt: We use lambda-case (`\case`) in other functions (like `mkSqliteConditions`) instead of a `case-of` so I think it would be better to stick to it (or to rewrite the other to use `case-of` as well).

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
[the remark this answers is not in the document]
eyeinsky: If the type signature is monomorphic then being indirect makes it less clear, no? We had this discussion in the Utxo indexer where a direct implementation was replaced with one depending on Monad and Alternative instances. Also, how do you choose what to write if there were competing polymorphic ways to write the same thing, are they all better than the direct version?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
eyeinsky: Done here 8c7e260f76bac84d4ba5ab56e197c909fb069da9

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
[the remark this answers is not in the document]
berewt: My reasoning is that one makes the structure explicit, while the other makes the purpose explicit. In one case, we insist on the fact that we return a pair of empty lists, in the other, we insist on the fact that it's the neutral element of this monoid. As the pair of lists isn't the nicer structure to read (to me), I prefer the latter.
But choosing between the two is definitely a matter of taste.
For example, I would probably prefer `""` over `mempty`, and definitely prefer `mempty` over `MkComplexDatatype [] "" 0 0 0` to give two extreme cases.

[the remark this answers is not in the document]
eyeinsky: Right, but since the return type is monomorphic then there is no "Monoid purpose", it's just that the type class definitions happen to match the current behavior. If the return type were `Monoid a => ... -> a`, then sure. With this (and the Utxo case as well) I'm arguing that using type class instances in the body of the function only adds indirection for no reason.

marconi-sidechain/src/Marconi/Sidechain/Api/Query/Indexers/MintBurn.hs
```diff
@@ -72,7 +72,6 @@ queryByPolicyAndAssetId env policyId assetId slotNo = do
             (x ^. txMintRowBlockHeaderHash)
             (x ^. txMintRowSlotNo)
             (x ^. txMintRowTxId)
-            Nothing
+            (Just $ C.hashScriptDataBytes $ C.unsafeHashableScriptData $ x ^. txMintRowRedeemerData)
```
koslambrou, line 75: Maybe not a good idea to use unsafeHashableScriptData.

I think you need to change the MintBurn indexer to work with `HashableScriptData` instead of just `ScriptData`.

eyeinsky: Could we leave this for a separate ticket? As I'm looking into this then I think this can't be done within cardano-api types as HashableScriptData is made by serialising ScriptData and in the conversion the serialised bytes for the script are already likely not the correct. (This is what I understand the [comment for `unsafeHashableScriptData`](https://github.com/input-output-hk/cardano-api/blob/cb74de6ab4006406003a8f11a24bfb83a32929dc/cardano-api/internal/Cardano/Api/ScriptData.hs#L118-L121) in cardano-api to say)

It seems that the way to get the hash is get it from the ledger transaction within the indexer _before_ it's converted into an cardano-api type [within the getPolicyData](https://github.com/input-output-hk/marconi/blob/7e1b044eab8ba9501a4b3bab9c19695c9926a88c/marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs#L112-L133) (maybe even cardano-ledger-api could help us here, but don't know enough about that).

There are uses of `unsafeHashableScriptData` in the Utxo indexer (and also in the ScriptTx indexer which is where it all probably started. Sorry about that :p) -- when we have the solution we might be able to fix all of them.

eyeinsky: Right, so the problem with the other `unsafeHashableScriptData` uses is that in the tests all `ScriptData`'s are not parsed from ledger script data but generated directly by the generators and there are no "original" bytes to hash.

PR #50, opened 2023-06-01
PLT-6068 Improve Utxo indexer performance

berewt:
Reverting the SQL representation of UTXO to two tables (unspent/spent). Add an index on TxId/TxIx to speed up the queries.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
@@ -257,8 +257,8 @@ data UtxoRow = UtxoRow
 
 $(makeLenses ''UtxoRow)
 
-urSpendSlotNo :: Traversal' UtxoRow C.SlotNo
-urSpendSlotNo = urSpentInfo . _Just . siSpentPoint . cpSlotNo
+urSpentSlotNo :: Traversal' UtxoRow C.SlotNo
+urSpentSlotNo = urSpentInfo . _Just . siSpentPoint . cpSlotNo
 
 urCreationSlotNo :: Lens' UtxoRow C.SlotNo
 urCreationSlotNo = urCreationPoint . cpSlotNo
@@ -295,7 +295,7 @@ instance ToJSON UtxoRow where
     [ "utxo" .= view urUtxo ur
     , "slotNo" .= view urCreationSlotNo ur
     , "blockHeaderHash" .= view urCreationBlockHash ur
-    , "spentSlotNo" .= preview urSpendSlotNo ur
+    , "spentSlotNo" .= preview urSpentSlotNo ur
     , "spentBlockHeaderHash" .= preview urSpentBlockHash ur
     , "spentTxId" .= preview urSpentTxId ur
     ]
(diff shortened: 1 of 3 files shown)


marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -186,7 +183,7 @@ data Utxo = Utxo
 $(makeLenses ''Utxo)
 
 instance Ord Utxo where
-  compare u u' = compare (u ^. address, u ^. txIn) (u' ^. address, u' ^. txIn)
+  compare u1 u2 = compare (u1 ^. txIn) (u2 ^. txIn)
```
kayvank, line 186: not comparing the address?

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -252,23 +255,22 @@ data UtxoRow = UtxoRow
   , _urSpentInfo     :: !(Maybe SpentInfo)
   } deriving (Show, Eq, Ord, Generic)
 
-
 $(makeLenses ''UtxoRow)
 
+urSpendSlotNo :: Traversal' UtxoRow C.SlotNo
```
kayvank, line 260: should probably be changed from `urSpendSlotNo` to `urSpentSlotNo` to remain consistent.  And replace other occurances of "\*spend\*" with "\*spent\*"

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -516,41 +529,49 @@ instance Buffered UtxoHandle where
...
-                          WHERE transactions_outputs.spentSlotNo IS NOT NULL|]
+        SQL.execute_ c [r|DELETE FROM
+                            unspent_transactions
+                          WHERE
+                            unspent_transactions.rowid IN (
+                              SELECT
+                                unspent_transactions.rowid
+                              FROM
+                                unspent_transactions
+                                JOIN spent ON unspent_transactions.txId = spent.txId
```
kayvank, line 572: when we left join the spent & unspent_transactions tables, why do we need to join them on the (txId, txIx)? rather than just txId

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -465,40 +474,44 @@ open dbPath (Depth k) isToVacuume = do
...
+                      , slotNo INT NOT NULL
+                      , blockHash BLOB NOT NULL
+                      , txIndexInBlock INT NOT NULL )|]
+
+    lift $ SQL.execute_ c [r|CREATE TABLE IF NOT EXISTS spent
+                      ( txId TEXT NOT NULL
+                      , txIx INT NOT NULL
+                      , slotNo INT NOT NULL
+                      , blockHash BLOB NOT NULL
+                      , spendTxId TEXT NOT NULL)|]
```
kayvank, line 496: change occurrences of `spendTxId` to spentTxId`

kayvank (approved):

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
berewt: I think we've discussed it and conclude that the `txIn` would be enough to differentiate Utxos. I honestly like to drop this ord instance and to use `compareOn` when needed, and use `Set` instead of `Map` when needed, but it can be done in another story.

berewt: I think we've discussed it and conclude that the `txIn` would be enough to differentiate Utxos. I honestly like to drop this ord instance and to use `compareOn` when needed, and use `Set` instead of `Map` when needed, but it can be done in another story.

PR #58, opened 2023-06-07
Utxo-Indexer optionally stores Reference script

kayvank:
Add CLI parameter to the Utxo indexer  to optionally enabling persisting the  ReferenceScript, inlineScript and the corresponding hash in the Utxo SQLite table, unspentTransactions.

- For sidechain, this parameter will be set to `False` indicating we do not save ReferenceSctips.

- For marconi-chain-index, this parameter by default is set to Enabled, saving the ReferenceScript. Users may turn it off using CLI parameter.  

For additional detail, see [PLT-5571](https://input-output.atlassian.net/browse/PLT-5571)

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
@@ -167,7 +167,7 @@ utxoWorker_
   -- ^ callback function used in the queryApi thread, needs to be non-blocking
   -> Utxo.Depth
   -> UtxoIndexerConfig
-  -- ^ utxo indexer configuration
+  -- ^ Utxo Indexer Configuration, containing targetAddresses and showReferenceScript flag
   -> Coordinator
   -> TChan (ChainSyncEvent (BlockInMode CardanoMode))
   -> FilePath
@@ -198,7 +198,7 @@ utxoWorker
   :: (Utxo.UtxoIndexer -> IO ())
   -- ^ callback function used in the queryApi thread
   -> UtxoIndexerConfig
-  -- ^ utxo indexer configuration
+  -- ^ Utxo Indexer Configuration, containing targetAddresses and showReferenceScript flag
   -> Worker
 utxoWorker callback utxoIndexerConfig securityParam coordinator path = do
   workerChannel <- atomically . dupTChan $ _channel coordinator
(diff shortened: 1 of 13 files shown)


marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -457,23 +459,33 @@ propUtxoQueryByAddressAndSlotInterval = property $ do
             raiseException $
               Utxo.open ":memory:" (Utxo.Depth 1) False
                 >>= Storable.insertMany events
-        let
+        let {-
```
kayvank, line 462: @berewt fourmolu insists on moving the comments through out this section of the code, a lambda function, to this location.

kayvank: I think the issue might me emacs related and related to this variable I set to use fourmolu:
`lsp-haskell-formatting-provider`

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -166,11 +198,13 @@ utxoWorker_
   -> Utxo.Depth
   -> Maybe TargetAddresses
   -- ^ Target addresses to filter for
+  -> Bool
+  -- ^ Store ReferenceScript
```
berewt, line 202: It may be worth creating a `UtxoIndexerConfig` type to group both options.
(and to pass it along as soon as we need to pass both parameters)

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -300,7 +302,8 @@ instance FromJSON UtxoRow where
             (Just s', Just bh', Just txId') -> Just $ SpentInfo (ChainPointRow s' bh') txId'
             _error -> fail "Inconsistent spent info"
      in UtxoRow
-          <$> v .: "utxo"
+          <$> v
+            .: "utxo"
```
berewt, line 306: weird linebreak... was it introduced by fourmolu?

berewt (changes requested):
A few clean up and it should be ok.
Can you add a test to the Utxo indexer with the reference script disabled (ensure that all the utxos have `Nothing` as a reference script).

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
koslambrou: I got the same thing, in VIM but only when using formoulu with HLS. Using `fix-fourmolu` from the CLI didn't introduce this weird linebreak.

berewt (approved):
Ok to ship once it passes the CI.

PR #63, opened 2023-06-14
PLT-5770 Updated the JSON-RPC request and response interface of `marconi-sidechain` to reflect the specification in API.adoc.

koslambrou:
Updated the JSON-RPC request and response interface of `marconi-sidechain` to reflect the specification in API.adoc.

* Added indexing of `BlockNo` in the AddressUtxo indexer
* Renamed the `blockHash` field in the AddressUtxo table to `blockHeaderHash` instead in order to be more explicit.
* Modified the `LastSyncQuery` to `LastSyncedBlockInfoQuery` which returns a `BlockInfo` type instead of `ChainPoint`.
* Fixed bug when starting `marconi-sidechain` and trying to query the MintBurn indexer before the indexer was able to index some information. In that scenario, the indexer didn't have time to be created, thus we got returned a query error. Instead of a query error, we just return an empty result.
* Updated the `marconi-sidechain` with up-to-date examples.
* Updated the `marconi-sidechain` golden test files for each JSON-RPC method.
* Added some post-conditions in the `marconi-sidechain` API specification

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/json-rpc/src/Network/JsonRpc/Types.hs
+++ b/marconi-chain-index/json-rpc/src/Network/JsonRpc/Types.hs
@@ -25,12 +25,17 @@ module Network.JsonRpc.Types (
   JsonRpcErr (..),
   JsonRpcResponse (..),
 
-  -- ** Standard error codes
+  -- ** Standard error codes and messages
   parseErrorCode,
+  parseErrorMessage,
   invalidRequestCode,
+  invalidRequestMessage,
   methodNotFoundCode,
+  methodNotFoundMessage,
   invalidParamsCode,
+  invalidParamsMessage,
   internalErrorCode,
+  internalErrorMessage,
 
   -- * Type rewriting
   JsonRpcEndpoint,
@@ -111,17 +116,32 @@ data JsonRpcErr e = JsonRpcErr
 parseErrorCode :: Int
 parseErrorCode = -32700
 
+parseErrorMessage :: String
+parseErrorMessage = "Parse error"
+
 invalidRequestCode :: Int
 invalidRequestCode = -32600
+
+invalidRequestMessage :: String
+invalidRequestMessage = "Invalid request"
 
 methodNotFoundCode :: Int
 methodNotFoundCode = -32601
 
+methodNotFoundMessage :: String
+methodNotFoundMessage = "Method not found"
+
 invalidParamsCode :: Int
 invalidParamsCode = -32602
 
+invalidParamsMessage :: String
+invalidParamsMessage = "Invalid params"
+
 internalErrorCode :: Int
 internalErrorCode = -32603
+
+internalErrorMessage :: String
+internalErrorMessage = "Internal error"
 
 instance (FromJSON e, FromJSON r) => FromJSON (JsonRpcResponse e r) where
   parseJSON = withObject "Response" $ \obj -> do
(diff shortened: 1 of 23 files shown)


marconi-chain-index/json-rpc/src/Network/JsonRpc/Types.hs
```diff
@@ -111,18 +116,33 @@ data JsonRpcErr e = JsonRpcErr
 parseErrorCode :: Int
 parseErrorCode = -32700
 
+parseErrorMessage :: String
+parseErrorMessage = "Parse error"
```
berewt, line 120: Don't we want to have a datatype that put together the error and the message?

marconi-sidechain/src/Marconi/Sidechain/Api/Query/Indexers/MintBurn.hs
```diff
@@ -63,7 +63,9 @@ queryByPolicyAndAssetId env policyId assetId slotNo = do
       tryReadTMVar $
         env ^. sidechainEnvIndexers . sidechainMintBurnIndexer . mintBurnIndexerEnvIndexer
```
berewt, line 64: I didn't pay attention to it in the past. It sounds like the query will fail if the indexer is busy indexing stuff. Do we really want this?

berewt (approved):
LGTM. One of the comment is unrelated but we may raise an issue for it.

marconi-sidechain/src/Marconi/Sidechain/Api/Query/Indexers/MintBurn.hs
koslambrou: It's not going to fail anymore. I removed the `Left ...`, and instead made it  return a `Right []` (empty result).

marconi-chain-index/json-rpc/src/Network/JsonRpc/Types.hs
koslambrou: Will do that.

marconi-sidechain/src/Marconi/Sidechain/Api/Query/Indexers/MintBurn.hs
berewt: Isn't it even worse? We return an empty result just because the indexer is "busy" (ie. the `MVar` content was taken by another thread).

marconi-chain-index/json-rpc/src/Network/JsonRpc/Types.hs
koslambrou: Can you validate the change? 43cdc8197e76de8d09ed49117fe71523ad821801

berewt: Even better.

marconi-sidechain/src/Marconi/Sidechain/Api/Query/Indexers/MintBurn.hs
koslambrou: > Isn't it even worse?

Maybe yeah. I just made it consistent with the Utxo indexer. 

Do we want to use `takeTMVar` instead, which retries until it succeeds as a temp solution?

berewt: Maybe better yes, I prefer the indexer to hang rather than returning inconsistent results.

koslambrou: I'll change it then

koslambrou: After changing it, it requires a bit of changes in `Marconi.ChainIndex.Indexers`. Therefore, I'll do in a separate PR.

PR #64, opened 2023-06-16
PLT-6283 Handle simple scripts

berewt:
This PR adds support of simple scripts, which don't have a redeemer. It makes the redeemer optional and doesn't bypass a minting that has no redeemer.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
```diff
@@ -85,18 +90,27 @@ data TxMintEvent = TxMintEvent
...
+  deriving (Eq, Ord, Show, Generic, SQL.FromRow, SQL.ToRow)
+
 data MintAsset = MintAsset
   { mintAssetPolicyId :: !C.PolicyId
   , mintAssetAssetName :: !C.AssetName
   , mintAssetQuantity :: !C.Quantity
-  , mintAssetRedeemerIdx :: !Word64
-  , mintAssetRedeemerData :: !C.ScriptData
-  , mintAssetRedeemerHash :: !(C.Hash C.ScriptData)
+  , mintAssetIdx :: !Word64
```
koslambrou, line 103: What is that field exactly? If there is no redeemer, then I wouldn've thought this would go with `MintAssetRedeemer`.

koslambrou (commented):
You also need to change the JSON-Schema in `API.adoc` and specify that the `redeemer` and `redeemerHash` fields in the response not required

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
berewt: It's the position of the asset in the list of assets. I don't think it's used anywhere. I kept it for legacy purposes but we can trash it.

PR #65, opened 2023-06-16
Replace tryReadTMVar with readTMVar when querying the indexers in marconi-sidechain.

koslambrou:
With tryReadTMVar, if the read failed because the indexer was not available yet, it would return an empty result. Now, with readTMVar, it restarts until the indexer is available after the write action has finished.

Doing so, we also needed to make sure that each indexer worker calls the callback when the indexer is initially created.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/EpochState.hs
+++ b/marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/EpochState.hs
@@ -17,13 +17,12 @@ import Control.Concurrent qualified as STM
 import Control.Concurrent.Async qualified as IO
 import Control.Concurrent.STM qualified as STM
 import Control.Exception (catch)
-import Control.Monad (forever, void, when)
+import Control.Monad (void)
 import Control.Monad.IO.Class (liftIO)
 import Data.Aeson ((.=))
 import Data.Aeson qualified as J
 import Data.ByteString.Lazy qualified as BL
 import Data.List qualified as List
-import Data.Map qualified as Map
 import GHC.Stack (HasCallStack)
 import Hedgehog qualified as H
 import Hedgehog.Extras.Test qualified as HE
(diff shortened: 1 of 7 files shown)


marconi-chain-index/test/Spec/Marconi/ChainIndex/Indexers/EpochState.hs
```diff
@@ -163,14 +162,15 @@ test = integration . HE.runFinallies . TN.workspace "chairman" $ \tempAbsPath ->
...
   found <- liftIO IO.newEmptyMVar
-  liftIO $ (IO.link =<<) $ IO.async $ forever $ do
-    (_, event) <- IO.readChan indexedTxs
-    case Map.lookup poolVKey (EpochState.epochStateEventSDD event) of
-      Just lovelace ->
-        when (lovelace == totalStakedLovelace) $
-          IO.putMVar found (EpochState.epochStateEventEpochNo event) -- Event found!
-      Nothing ->
-        return ()
+  -- TODO Uncomment once we support notifications.
```
koslambrou, line 165: @berewt Reason I commented this:

In order to support this part, I need the callback to be `(State h, StorableEvent h) -> IO ()`. However, I had to change the callback of `EpochState` to just `State h -> IO ()`, because we need to call that callback when a rewind takes place, thus we have no `StorableEvent` to give.

Alternative, we can change the callback to `(State h, Maybe (StorableEvent h)) -> IO ()`, but feels adhoc. Maybe I can do that temporarily. As we're thinking of supporting notifications, this would be a good scenario here.

What's more, this test is never run in our CI, because of some issue with `cardano-testnet`. So we have no idea if this test actually works.

What do you think?

koslambrou: So I'll keep this part commented?

koslambrou:
@berewt Approve?

PR #66, opened 2023-06-16
Replaced ChainPoint in UtxoEvent with BlockInfo

koslambrou:
In the UtxoEvent, we used to have `ChainPoint` which could be `ChainPointAtGenesis`. However, it should be impossible to create an event which is at genesis. Therefore, we can just use `BlockInfo` instead which contains additional info such as `BlockNo`, `BlockTimestamp` and `EpochNo`. 

This change also means that we don't need to handle the `Genesis` case in most of the code, thus we gain some nice refactoring.

@berewt Part I've contemplated is if we should put the `BlockInfo` in a separate table, and then the `utxo` table and the `spent` table would point to the correct row in the `BlockInfo` table. Thoughts?

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
@@ -61,7 +61,7 @@ import Data.Aeson (
   (.=),
  )
 import Data.Either (fromRight)
-import Data.Foldable (fold, foldl', toList)
+import Data.Foldable (foldl', toList)
 import Data.List (groupBy, sortOn)
 import Data.List.NonEmpty (NonEmpty)
 import Data.Map (Map)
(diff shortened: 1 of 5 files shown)


marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -259,6 +255,24 @@ instance ToJSON Utxo where
...
+  , _blockInfoBlockNo :: !C.BlockNo
+  , _blockInfoTimestamp :: !Word64
+  , _blockInfoEpochNo :: !C.EpochNo
+  }
+  deriving (Eq, Show, Ord)
+
+$(makeLenses ''BlockInfo)
+
+instance FromRow BlockInfo where
+  fromRow = BlockInfo <$> field <*> field <*> field <*> field <*> field
```
berewt, line 270: We can derive this one.

berewt (approved):
LGTM

PR #67, opened 2023-06-17
Reduce the need of `Point` type family

berewt:
`TimedEvent` is great to attach a chainPoint to a datatype, it was hard to manipulate though, because each type should have a `Point` type instance to use it, and we couldn't map on it because of the type family.

- Change `TimedEvent event` to `TimedEvent point event`.
- Add `Functor`, `Foldable` and `Traverse` to the experimental indexer.
- Refactor code to take advantage of it

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Utxo.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Utxo.hs
@@ -48,7 +48,6 @@ import Database.SQLite.Simple.ToField (ToField (toField))
 import Database.SQLite.Simple.ToRow (ToRow (toRow))
 import Text.RawString.QQ (r)
 
-import Cardano.Api (AddressAny)
 import Cardano.Api qualified as C
 import Cardano.Api qualified as Core
 import Cardano.Api.Shelley qualified as C
@@ -77,12 +76,9 @@ data Utxo = Utxo
 $(makeLenses ''Utxo)
 
 instance Ord Utxo where
-  compare u1 u2 =
-    compare (u1 ^. utxoTxIn) (u2 ^. utxoTxIn)
+  compare u1 u2 = compare (u1 ^. utxoTxIn) (u2 ^. utxoTxIn)
 
-newtype Spent = Spent
-  { unSpent :: C.TxIn
-  }
+newtype Spent = Spent {unSpent :: C.TxIn}
   deriving (Show, Eq, Ord)
 
 data UtxoEvent = UtxoEvent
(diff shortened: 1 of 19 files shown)


marconi-chain-index/test/Spec/Marconi/ChainIndex/Experimental/Indexers/Utxo/UtxoIndex.hs
```diff
@@ -85,14 +85,6 @@ tests =
...
     ]
 
--- Fro test purpose only. We do not address ChainPoint, `Point`
--- We do not address Ord or Eq of `Point`. Tests are such that Points are constant.
-instance Eq e => Eq (Core.TimedEvent e) where
-  u == u' = u ^. Core.event == u' ^. Core.event
-
-instance Ord e => Ord (Core.TimedEvent e) where
-  u `compare` u' = (u ^. Core.event) `compare` (u' ^. Core.event)
-
```
berewt, line 95: Moved to `Marconi-core`

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Utxo.hs
```diff
@@ -93,16 +94,14 @@ data UtxoEvent = UtxoEvent
 $(makeLenses ''UtxoEvent)
 
 type instance Core.Point UtxoEvent = C.ChainPoint
-type instance Core.Point Utxo = C.ChainPoint -- required for the unspent_transaction table SQL mappings
-type instance Core.Point Spent = C.ChainPoint -- required for the spent table SQL mappings
```
berewt, line 97: This is why I wanted to make the change: having to declare a type instance for each type that you need to decorate with a `Point` was tedious.

koslambrou: As the `Point` type family still seems to be used in `marconi-core`, how do you map `ChainPoint` to `Point event`? For example, I see `(TimedEvent (Point event) event)` in some signatures.

berewt: We still have to declare it for `event`, but not for intermediary types that are used internally when we're processing them probably `TimedData` would be a better name for it now.
It allows some simplification in the indexer implementations, [like this one](https://github.com/input-output-hk/marconi/pull/67/files#diff-c07cb25c3219e34c21ad277adaed60d1df737d45eba462fa7ced76487f1b4ef9R174).

koslambrou: I understand now. Thanks for the clarification :)

koslambrou (approved):

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Utxo.hs
berewt: Do you think it worths renaming `TimedEvent` to `Timed` or `TimedData` before merging it?

PR #73, opened 2023-06-22
ADR on marconi observability

koslambrou:
* Added draft ADR for how we observe the runtime behaviour of Marconi
* Added script in Nix shell (disabled by default) which starts a VM
  running Grafana and Prometheus

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/doc/read-the-docs-site/adr/0007-observability.rst
+++ b/doc/read-the-docs-site/adr/0007-observability.rst
@@ -1,0 +1,49 @@
+.. _adr7:
+
+ADR 7: Marconi observability
+============================
+
+Date: 2023-07-19
+
+Authors
+-------
+
+koslambrou <konstantinos.lambrou@iohk.io>
+
+Status
+------
+
+Draft
+
+Context
+-------
+
+Whenever we add new features in Marconi, we inevitably end up in a state where we introduce a significant regression in sync time, CPU/RAM/Disk usage and query performance.
+Currently, developers who are adding new features will typically test those features by running ``marconi-chain-index`` or ``marconi-sidechain``, and manually visualize the CPU/RAM usage by running ``htop`` (or similar), checking syncing speed by reading the logs, and manually run queries after syncing is done.
+However, that approach does not give accurate information for detecting regressions, because we only look at the logs at specific points in time and we might miss peaks in memory usage for example.
+Moreover, this testing approach does not scale for delivering on time the different features.
+
+Therefore, we need a solution for observing the runtime behavior of Marconi through different metrics such blocks processed, query time, CPU/RAM/Disk usage, etc.
+Ideally, these observations would be done through graphical visualizations.
+Eventually, they would be used for external reporting and as part of our QA practices.
+
+Decision
+--------
+
+* We will expose Prometheus metrics for our chain-indexers, more specifically, ``marconi-sidechain``.
+  The metrics we should provide are:
+
+  * CPU, memory and disk usage
+  * Total number of processed blocks
+  * The number of processed queries
+  * The time spent processing queries
+
+  Except the CPU/RAM/Disk usage which are provided by the systemd process, the application-specific metrics will be exposed in an HTTP endpoint such as ``/metrics``.
+
+* We will expose Grafana dashboards to visualize each of the metrics.
+
+Argument
+--------
+
+Grafana and Prometheus seem to be the go-to framekwork for observing behavior of long-running applications, and it is used by other projects in IOG, most notably by the Sidechain Tribe.
+As we're delivering ``marconi-sidechain`` for them, it makes sense to use the same framework.
--- a/flake.nix
+++ b/flake.nix
@@ -7,7 +7,9 @@
 
   inputs = {
 
-    iogx.url = "github:input-output-hk/iogx";
+    # TODO The `?ref=remove-list-binaries` is a temporary solution to solve a CI error.
+    # Probably an # edge case in IOGX.
+    iogx.url = "github:input-output-hk/iogx?ref=remove-list-binaries";
     iogx.inputs.CHaP.follows = "CHaP_2";
 
     CHaP_2 = {
@@ -18,7 +20,9 @@
     # Used to provide the cardano-node and cardano-cli executables.
     cardano-node = {
       url = "github:input-output-hk/cardano-node";
-      flake = false;
+    };
+    mithril = {
+      url = "github:input-output-hk/mithril";
     };
   };
 
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -1,0 +1,12 @@
+# Benchmarking
+
+Steps on running `prometheus` and `grafana` locally:
+
+* Change the `enable` of `start-benchmark-machine` in `shell.nix` to `true`
+* Open a Nix shell with `nix develop`
+* Run `cardano-node`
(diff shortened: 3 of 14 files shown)


doc/read-the-docs-site/adr/0007-observability.rst
```diff
@@ -0,0 +1,63 @@
...
+    * Save the configuration on version control.
+    * Setup up manual deployment of configuration with ``Morph``.
+    * Setup up continuous deployment of configuration with ``Morph`` for each commit on the ``main`` branch of ``marconi-ops``.
+
+* We will add cardano-node preview, preprod and mainnet services which continuously run in the benchmarking machine.
+
+* We will configure Grafana on the benchmarking machine and make it publicly available for visualizing
+  the different Prometheus metrics exposed by ``marconi-sidechain``.
+
+* We will configure a *daily* cron job which will run ``marconi-sidechain`` on the latest commit of the ``main`` branch.
```
james-iohk, line 53: I'm currently running `marconi-sidechain` with multi threads and at 79% after 24 hours. That's with an already synced cardano-node.

1. Do we know the benchmarking machine will always be capable of syncing within 24 hours for a daily job?
2. There definitely won't be enough time to run consecutive syncs on each of the 3 networks daily so we may need to come up with a more strategic schedule.

doc/read-the-docs-site/adr/0007-observability.rst
```diff
@@ -0,0 +1,63 @@
...
+  These metrics will be exposed in an HTTP endpoint such as ``/metrics``.
+
+* We will bootstrap a "benchmarking" machine for the team.
+    * Create a private repository named ``marconi-ops``.
+    * Create an AWS instance based of a NixOS image and give the main engineers SSH access.
+    * Save the configuration on version control.
+    * Setup up manual deployment of configuration with ``Morph``.
+    * Setup up continuous deployment of configuration with ``Morph`` for each commit on the ``main`` branch of ``marconi-ops``.
+
+* We will add cardano-node preview, preprod and mainnet services which continuously run in the benchmarking machine.
```
james-iohk, line 48: Will the cardano-node instances all be running simultaneously to keep fully synced at all times?
If so, that could have an impact on benchmarking results when real-world instances would typically use a single network (unless there is some isolation).

berewt:
We may want to monitor the query side as well. eg. removing all the SQL indexes would increase the performance on the sync side, but would drastically degrade the query performances.

doc/read-the-docs-site/adr/0007-observability.rst
```diff
@@ -0,0 +1,63 @@
...
+Currently, developers who are adding new features will typically test those features by running ``marconi-sidechain`` on the Cardano preprod network.
+However, that network is very small compared to the Cardano mainnet, and cannot adequatly detect sync time and memory usage regressions.
+Morever, developers are not always in the position to test their changes on mainnet because sync time typically takes around 20h, and they need a computer with at least around 32GB of RAM (~12GB for ``marconi-sidechain`` +  ~12GB for ``cardano-node`` + other processes).
+This large sync times and memory usage are typically because of the ``EpochState`` indexer which uses a large amount of memory because of the ``LedgerState`` computation.
+Even if we manage to remove the ``LedgerState`` computation as planned, sync time is still expected to be hours long.
+
+What we want is a dedicated server which runs the ``marconi-sidechain`` executable daily and exposes different metrics in order to detect any regressions in sync time, CPU, memory and disk usage throught graphical visualizations.
```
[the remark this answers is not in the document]
james-iohk: could we do it for a slice of mainnet to use for comparison? perhaps the Alonzo hype, e.g. a few months around block [6535793](https://cardanoscan.io/block/6535793)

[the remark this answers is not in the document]
berewt: It would mean having an upper-bound limit in Marconi. It's a bit of effort but it should be doable.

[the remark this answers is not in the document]
james-iohk: The test could detect the upper limit and kill the process when it's stored the measurements

doc/read-the-docs-site/adr/0007-observability.rst
koslambrou: 1. That's our goal (syncing under 24h). 
2. In the end, maybe that's not that much an issue. Even if the sync on mainnet is not finished, we can make it such that we can visualize previous syncs.

Do you have a better suggestion?

doc/read-the-docs-site/adr/0007-observability.rst
koslambrou: But once synced, pre-prod and preview use very little resources, right? I feel that it wouldn't impact the machine that much.

doc/read-the-docs-site/adr/0007-observability.rst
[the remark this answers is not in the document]
koslambrou: Good idea.

doc/read-the-docs-site/adr/0007-observability.rst
james-iohk: Yes perhaps. Let's see :)

doc/read-the-docs-site/adr/0007-observability.rst
james-iohk: My only suggestion would be to not run daily until it can sync all 3 networks in time otherwise you'll get incomplete results or an overlap.

james-iohk:
Are we defining observability of query performance too here?
If not, we should have something defined. We could consider using Artillery (or similar tool) because out-the-box we get easily configure simulation of load ramping and number of users and generates reports for each test run with useful measurements. Unless there's something comparable available in Haskell?

koslambrou:
> Are we defining observability of query performance too here? If not, we should have something defined. We could consider using Artillery (or similar tool) because out-the-box we get easily configure simulation of load ramping and number of users and generates reports for each test run with useful measurements. Unless there's something comparable available in Haskell?

@james-iohk There is `tasty-bench` which we currently use to benchmark query performance over certain indexers. However, it just measures the amount of time spent doing a single query. Artillery assumes there are **many** different users and multiple requests for querying that same endpoint. For `marconi-sidechain`, the number of requests for each endpoint are pretty limited. For example, there should be a single user doing ~1query/s on the Utxo endpoint. 

You think Artillery still has it's place?

doc/read-the-docs-site/adr/0007-observability.rst
koslambrou: Another idea would be to have multiple machines, each running a different network :) That's an orchestration problem now :D

flake.nix
```diff
@@ -18,7 +18,9 @@
     # Used to provide the cardano-node and cardano-cli executables.
     cardano-node = {
       url = "github:input-output-hk/cardano-node";
-      flake = false;
+    };
+    mithril = {
+      url = "github:input-output-hk/mithril";
```
berewt, line 23: FMI: is that to speed up the node setup?

berewt (approved):
LGTM

flake.nix
koslambrou: That's right. Very fast node bootstrapping.

PR #78, opened 2023-06-26
PLT-6329 Add Value in GetUtxoFromAddress response

berewt:
Add sidechain-specific JSON instances for `Value` it's done thanks to a newtype wrapper around cardano-api `Value`. 

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


eyeinsky (approved):
The main change seems to be adding C.Value into AddressUtxoResult and that looks good.

There is also extra work here that we don't actually use (?): reading the SidechainValue from JSON. The looks like a wrapper around `serialiseToRawBytesHexText`/`deserialiseFromRawBytesHex` for PolicyId and AssetName, perhaps these instances are already tested elsewhere?

berewt:
> There is also extra work here that we don't actually use (?): reading the SidechainValue from JSON.

The `FromJSON` instance is used in the `FromJSON` instance of the `AdressUtxoResult`

> The looks like a wrapper around `serialiseToRawBytesHexText`/`deserialiseFromRawBytesHex` for PolicyId and AssetName, perhaps these instances are already tested elsewhere?

The tripping for raw bytes hex might be tested in Cardano-node, but we use it mainly to build keys for the JSON objects here, so the tests are more about the JSON structure than about the bytes serialisation itself. (I may have misunderstood the comment though)

berewt:
Which test would you suggest to ensure that a JSON representation contains the exact same information as its Haskell counterpart? It's fairly easy to miss a field / put the wrong value in one of them / encode thing poorly. I do agree that we could use a custom function in the test code instead of the FromJSON instance. It may be harder to maintain so, if we don't have a better way to test that the encoding is sound and complete, I'm fine with this compromise.

eyeinsky:
It's just that to me all this looks very low priority. It's like if you had a string somewhere `str = "some string"` and then you wrote a test to test if that string actually is `"some string"`.

PR #83, opened 2023-06-29
Clean exit and better resume points

berewt:
This PR has introduce two new features:

- We now wait for each indexer to process their block before exiting Marconi. The wait has a 3 minutes timeout.
- We keep track of all the chainpoints when we flush on disk. I wanted to keep only one event first, but as we flush the whole memory at the moment and as we want to be able to resume from at least 1 non volatile point, I had to do so.

This PR doesn't revert the intersection computation. It should be done if we want to continue with this indexer design.
It isn't a problem for side chain as the epoch state indexer will lag behind anyhow and will be the one that will be used when resuming.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/bench/BenchQueries.hs
+++ b/marconi-chain-index/bench/BenchQueries.hs
@@ -42,6 +42,7 @@ import Data.Word (Word64)
 import Database.SQLite.Simple (FromRow (fromRow), field, toRow)
 import Database.SQLite.Simple qualified as SQL
 import Database.SQLite.Simple.FromField (FromField)
+import Database.SQLite.Simple.QQ (sql)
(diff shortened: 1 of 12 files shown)


james-iohk:
do you know what the behaviour on forced exit is?

berewt:
Forced exit as in `kill -9`? Uncovered so it might be a mess if it happens in the middle of a flush or while writing the ledger state. `^C` behave as described in the PR description.

james-iohk:
or signal 15 as a system would on shutdown. Just wondered if would still attempt some recovery. Of course there is no guarantee.

koslambrou (approved):
LGTM :)

PR #84, opened 2023-06-30
Ml/plt 5384  mintburn dbsync compare

eyeinsky:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


eyeinsky:
@koslambrou @berewt So it looks like I won't be able to finish this ticket.

The issue that remains is that redeemer hash and data don't always match between marconi and dbsync (if you set it to Nothing on both marconi and dbsync results then the test succeeds). The funny part being that they _sometimes_ don't match. Might be worth it to double-check the sql query with dbsync team and also and maybe ledger or plutus-core team on whether the redeemer we save is correct.

There are `TODO` items in the test which help to see where it goes wrong.

How to run the test:
- sync mintburn indexer on preprod
- sync dbsync on preprod
- set env variables
- (run the test as described in issue description)

PR #103, opened 2023-07-18
PLT-6210 JSON error for the mint-burn indexer

berewt:
- Add specific error messages for the getBurnTokenEvent JSON-RPC endpoint.
- Restructure the Marconi-chain-index errors to introduce indexer specific errors.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Error.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Error.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE OverloadedStrings #-}
 
 module Marconi.ChainIndex.Error where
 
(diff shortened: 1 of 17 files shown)


marconi-chain-index/src/Marconi/ChainIndex/Error.hs
```diff
@@ -37,10 +36,22 @@ liftSQLError errorWrapper =
...
-raiseException
-  :: ExceptT IndexerError IO a -> IO a
+raiseException :: (Exception err) => ExceptT (IndexerError err) IO a -> IO a
 raiseException x = do
   x' <- runExceptT x
   case x' of
     Left err -> throw err
     Right result -> pure result
+
+ignoreQueryError :: ExceptT (IndexerError err) IO a -> ExceptT (IndexerError Void) IO a
```
koslambrou, line 46: By reading the code, it is not 100% clear to me why we need to use `Void`. Just for my own comprehension.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/MintBurn.hs
```diff
@@ -108,6 +115,17 @@ data MintAsset = MintAsset
   }
   deriving (Show, Eq, Ord)
 
+-- | Errors that can occurs when you query the indexer
+data MintBurnError
```
koslambrou, line 119: Change to MintBurnQueryError if they are errors specifically for querying?

koslambrou (approved):
LGTM :)

marconi-chain-index/src/Marconi/ChainIndex/Error.hs
berewt: Here is the rationale: each query has its own exception. Each indexer has one and one only `StorableMonad`, that must be able to handle these exceptions and the other indexer exceptions. To allow a specialisation by indexer, we provide a parametric `IndexerError` type.
But, when we deal with `Workers` in the `Indexers` type class, all indexers must have the same monad stack to manage block indexing in the same way. As during indexing query error can't occur, the easiest solution for this common monad stack is to claim that it can't occur, using `Void`.
So `ignoreQueryError` deals with the `QueryError` case but in practice, this case shouldn't appear.

koslambrou: Thanks :)

PR #112, opened 2023-07-25
Add a catch-up mechanism at an indexer transformer

berewt:
As we want to move towards full in-database indexers, we discussed the need to have a way to batch events to the database until we catchup with the tip.

This PR proposes a indexer transformer that does exactly this: when an block comes in, we compute the distance to the tip and we decide either:

- that we are too far from the tip and thus we put the block in a batch and if the batch is full, we send it to the database
- that we are close enough to the tip, we just skip the batch mechanism and add directly the block to the indexer.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marconi-core/src/Marconi/Core/Experiment/Transformer/WithCatchup.hs
+++ b/marconi-core/src/Marconi/Core/Experiment/Transformer/WithCatchup.hs
@@ -1,0 +1,177 @@
+{-# LANGUAGE ApplicativeDo #-}
+{-# LANGUAGE DerivingVia #-}
+
+{- | The catch-up mechanism buffered events to pass them in batch to an indexer, to reduce the time
+ needed for an indexer to catch-up with the tip.
+
+ Once we are close enough to the tip of the chain, `WithCatchup` deactivate itself and pass directly
+ the blocks to the indexer.
+-}
+module Marconi.Core.Experiment.Transformer.WithCatchup (
+  WithCatchup,
+  withCatchup,
+  HasCatchupConfig (..),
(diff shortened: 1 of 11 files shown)


marconi-core/src/Marconi/Core/Experiment/Transformer/WithCatchup.hs
```diff
@@ -0,0 +1,177 @@
...
+resetBuffer :: WithCatchup indexer event -> WithCatchup indexer event
+resetBuffer = (catchupBufferLength .~ 0) . (catchupBuffer .~ [])
+
+instance (IsIndex m event indexer) => IsIndex m event (WithCatchup indexer) where
+  index timedEvent indexer =
+    let bufferIsFull ix = (ix ^. catchupBufferLength) >= (ix ^. catchupBatchSize)
+        pushEvent ix =
+          ix
+            & catchupBuffer %~ (timedEvent :)
+            & catchupBufferLength +~ 1
```
koslambrou, line 143: You keep the length in memory for efficiency reasons?

koslambrou (approved):

marconi-core/src/Marconi/Core/Experiment/Transformer/WithCatchup.hs
berewt: yes, it isn't a huge perf penalty but I feel dirty when I go through several thousands of elements just to get a length.

PR #113, opened 2023-07-26
PLT-6559 Fixed bug in EpochState indexer

koslambrou:
* Removed the node tip block number in the LedgerState filename stored on disk
* Use only the LedgerState filenames resumable points instead of the points stored on the database for the EpochSDD and EpochNonce tables.
* When starting the application, query the current node tip and pass that info on each indexer. That info is necessary for EpochState indexer in order to remove old LedgerStates that have become immutable.
* Initially, we only stored immutable LedgerState at the beginning of each epoch. Now, we also store the immutable LedgerState right before a rollbackable point. This capability allows us to resume to a point closer to the tip when almost fully synced.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
@@ -216,7 +216,11 @@ initialCoordinator indexerCount' minIndexingDepth =
 -- is a list of points (rather than just one) since it offers more resume possibilities
 -- to the node (in the unlikely case there were some rollbacks during downtime).
 type Worker =
-  SecurityParam -> Coordinator -> FilePath -> IO (Storable.StorablePoint ScriptTx.ScriptTxHandle)
+  SecurityParam
+  -> C.BlockNo
+  -> Coordinator
+  -> FilePath
+  -> IO (Storable.StorablePoint ScriptTx.ScriptTxHandle)
 
 utxoWorker_
   :: (Utxo.UtxoIndexer -> IO ())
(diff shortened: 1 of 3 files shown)


marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -617,9 +627,11 @@ runIndexers
   -> IO ()
 runIndexers socketPath networkId cliChainPoint indexingDepth (ShouldFailIfResync shouldFailIfResync) traceName indexerList = do
   securityParam <- Utils.toException $ Utils.querySecurityParam @Void networkId socketPath
+  currentNodeBlockNo <- Utils.toException $ Utils.queryCurrentNodeBlockNo @Void networkId socketPath
```
berewt, line 630: as the `currentNodeBlockNo` is specific to one worker, can't we resolve it in `main` pass it to the `epochStateWorker` (using partial application), and keep the other indexer clean?

berewt (approved):
Looks fine but I suggest an alternative solution, it's up to you to decide whether you want to consider it or not.

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
koslambrou: I initially did that, but it thought that it "corrupted" the `main` with too much indexer specific logic. Then, I thought that this `BlockNo` _could_ be beneficial to other indexers. 

I don't like either of these solutions though, but the idea of keeping `main` clearer "felt" better. I got no objective argument here.

Given my statement, you still think it makes more sense to make the currentNodeBlockNo argument specific to the `EpochState` worker?

berewt: we could also consider passing the `networkId` and the `socketPath` to the `epochState` and make it monadic. Is it better?
If you still think that the current solution is the least bad one, I'm fine with it, we'll see if we also have to deal with it (and if we can find a better alternative) with the new indexers.

berewt: it will return an `IO Worker` instead of a `Worker`

koslambrou: Like we discussed, we will revisit this on the new API version.

PR #116, opened 2023-07-26
PLT-6816 Change decision of Marconi resuming ADR with new implementation

koslambrou:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
+++ b/doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
@@ -3,7 +3,7 @@
 ADR 6: Indexer resuming strategy
 ================================
 
-Date: 2023-06-07
+Date: 2023-07-26
 
 Authors
 -------
@@ -13,7 +13,7 @@ koslambrou <konstantinos.lambrou@iohk.io>
 Status
 ------
 
-Proposed
+Draft
 
 Context
 -------
@@ -45,16 +45,87 @@ Decision
   In the current Marconi interface, it is actually just ``securityParam + 1`` as we assume that all rollbackable blocks *can* be fully stored on disk because of the disk flush.
   We assume that all the data contained in the indexer is consistent for all the point returned by an indexer.
 
-* We will have each indexer start the node-to-client chain-sync protocol at different points in time.
-  However, the ``Coordinator`` will make sure that all indexers are synced until a common point, and advance at the same speed (i.e. they can only request the next block once all the indexers have finished processing the current block).
+* We will start a single node chain-sync client for all indexers instead of one node chain sync client per indexer.
 
-* We will use the full set of points provided by ``resumableFromStorage`` of an indexer as resuming points for the node-to-client chain-sync protocol.
-  These points need to ordered in descending order so that protocol can priotise the selection of the latest ones.
+* We will implement the following resuming logic.
+
+  In order to start the chain-sync client, we need to provide to provide intersection points (or resuming points).
+  Intersection points represent the points from which we want to start syncing from the local node.
+  We will provide as intersection points the points returned from the following pseudo-code:
+
+  .. code-block:: haskell
+
+    -- | Query the resumable points of each indexer then select the resumable
+    -- points of the indexer that has the lowest immutable point. If two indexer
+    -- have the same lowest immutable point, select the one which has the lowest
+    -- overall point (immutable or not).
+    selectIntersectionPoints indexers = do
+      resumingPoints <- mapM queryResumingPoints indexers
+      head $ sortOn (\rps -> (maximum $ filter isImmutablePoint rps, maximum rps)) resumingPoints
+
+  Next, we need a function which processes each ``ChainSyncEvent`` returned by the chain-sync client for each indexer.
+  The pseudo-code is as follows:
+
+  .. code-block:: haskell
+
+    applyChainSyncEvent previousChainSyncEventPoint event indexer = do
+      case event of
+        lastSyncPoints <- getLastSyncPoints indexer
+        Resume point -> do
+          maybePoint <- getPointFromSlotNo (slotNo point) indexer
+          case maybePoint of
+            Nothing ->
+              pure () -- Do nothing
+            Just _ ->
+              rewind point indexer
+        RollBackward point -> do
+          rewind point indexer
+        RollForward block point -> do
+          lastSyncPoints <- getLastSyncPoints indexer
+          let lastSyncSlotNos = fmap getSlotNo lastSyncPoints
+          let lowestLastSyncSlotNo = minimum lastSyncSlotNos
+          let highestLastSyncSlotNo = maximum lastSyncSlotNos
+
+          -- Block already processed. Do nothing
+          if point `elem` lastSyncPoints
+          then do
+            pure ()
+
+          -- SlotNo of new block point already processed, but with different block header hash
+          -- We need to rewind and insert the new block
+          else if getSlotNo point `elem` lastSyncSlotNos
+          then do
+            rewind previousChainSyncEventPoint indexer
+            insert block indexer
+
+          -- Block already processed. Do nothing
+          else if getSlotNo point < lowestLastSyncSlotNo
+          then do
+            pure ()
+
+          -- New block not processed yet. We insert it in the indexer
+          else if getSlotNo point > highestLastSyncSlotNo
+          then do
+            insert block indexer
+
+          -- Should never go here
+          else do
+            pure ()
+
+  It is important to note that the implementation of ``applyChainSyncEvent`` highly depends on the implementation of ``selectIntersectionPoints``.
 
 Argument
 --------
 
-In order to justify the decision, we will present various use case scenarios and show how the decisions satisfies them.
+The reason ``resumeFromStorage`` returns a list of resuming points where there is *at least* one immutable point is to ensure we can resume from an existing chain point when restarting Marconi.
+Then, the reason why we include rollbackable points in the list of resuming points is simply to resume from a point close to the tip, thus prevent re-indexing of already indexed information.
+
+The main drawback of the solution is that each indexer must keep track of the last sync points (although all solutions written thus far relied on this assumption).
+We provide a way to create the table implicitly when using the SQLite adapter.
+However, if the user wants to use a different db, then he will need to rewrite the lastSyncPOints logic in the database.
+However, we assume that most user will want to use the provided adapters, and not write indexers in databases that are not supported yet.
+
+In order to justify the decision, we will present various use case scenarios and show how the new implementation works.
 We assume two indexers: ``A`` and ``B`` which have started indexing information, and then were **abruptly** stopped.
 The use cases will show what will happen when resuming them.
 
@@ -73,30 +144,14 @@ In Haskell, that would look something like:
 
 ::
 
-              Rollbackable
-            |--------------|
+                   Rollbackable
+                     |-----|
   1 2 3 4 5 6 7 8 9 10 11 12
     |---|                  |
       A                   Tip
 
 ``A``'s resumable interval provided for the chain-sync protocol is ``[4, 3, 2]``.
-The chain-sync protocol is started at point ``4``, thus ``A`` is rewinded to point ``4``.
-
-``A`` has a resumable interval fully included in the rollbackable chain point interval
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-::
-
-                 Rollbackable
-                |-------------|
-  1 2 3 4 5 6 7 8 9 10 11 12 13
-                |----|        |
-                   A         Tip
-
-``A``'s resumable interval provided for the chain-sync protocol is ``[10, 9, 8]``.
-The chain-sync protocol will try each of these points and identify the first one which is known by the local node.
-As rollbacks can occur between points ``[8..13]`` after the indexer was stopped, the points ``[8..9]`` provided by the indexer *could* be invalid.
-Thus, if any of those points fail, the chain-sync protocol will just start from genesis.
+The chain-sync protocol is started at point ``4``, thus ``A`` starts syncing from point ``5`` with no lost data.
 
 ``A`` has a resumable interval overlapping the rollbackable chain point interval
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
```diff
@@ -45,16 +45,88 @@ Decision
...
+
+  .. code-block:: haskell
+
+    -- | Query the resumable points of each indexer then select the resumable
+    -- points of the indexer that has the lowest immutable point. If two indexer
+    -- have the same lowest immutable point, select the one which has the lowest
+    -- overall point (immutable or not).
+    selectIntersectionPoints indexers = do
+      resumingPoints <- mapM queryResumingPoints indexers
+      head $ sortOn (\rps -> (maximum $ filter isImmutablePoint rps, maximum rps)) resumingPoints
```
berewt, line 64: Is this correct? It doesn't look like an intersection. Actually do we want to return an intersection at all? ie. if one indexer is behind, the intersection is empty but we want to restart from there. I don't know what would be the correct set of points to send... all the points from the least advanced indexer?

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
```diff
@@ -45,16 +45,88 @@ Decision
...
+          maybePoint <- getPointFromSlotNo (slotNo point) indexer
+          case maybePoint of
+            Nothing ->
+              pure () -- Do nothing
+            Just _ ->
+              rewind point indexer
+        RollBackward point -> do
+          rewind point indexer
+        RollForward block point -> do
+          lastSyncPoints <- getLastSyncPoints indexer
```
berewt, line 84: we may want to avoid recomputing it after each new event.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
```diff
@@ -45,16 +45,88 @@ Decision
...
+          if point `elem` lastSyncPoints
+          then do
+            pure ()
+
+          -- SlotNo of new block point already processed, but with different block header hash
+          else if getSlotNo point `elem` lastSyncSlotNos
+          then do
+            -- TODO Not exactly correct, we need to rewind to `point - 1`, but we don't necessarly have
+            -- the block header hash of that point
+            rewind point indexer
```
berewt, line 99: it doesn't include the idea of tracking the point of the last traced event we discussed, isn't it?

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
```diff
@@ -45,16 +45,88 @@ Decision
...
+          -- New block not processed yet.
+          else if getSlotNo point > highestLastSyncSlotNo
+          then do
+            insert block indexer
+
+          -- Should never go here
+          else do
+            pure ()
+
+  It is important to note that the implementation of ``applyChainSyncEvent`` is highly depend on the implementation of ``selectIntersectionPoints``.
```
berewt, line 116: is highly depend -> highly depends

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
```diff
@@ -45,16 +45,88 @@ Decision
...
 
 Argument
 --------
 
-In order to justify the decision, we will present various use case scenarios and show how the decisions satisfies them.
+The reason ``resumeFromStorage`` returns a list of resuming points where there is *at least* one immutable point is to ensure we can resume from an existing chain point when restarting Marconi.
+Then, the reason why we include rollbackable points in the list of resuming points is simply to resume from a point close to the tip, thus prevent re-indexing of already indexed information.
+
+The main drawback of the solution is that each indexer must keep track of the last sync points.
```
berewt, line 124: Add a note that all the solutions written so far relied on this assumption, although it wasn't always clearly stated.

berewt (approved):
The main paint point is the selection of the resumablePoints to send to the chainSync client.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: Sorry, the name is wrong. It's intersection with the points from the local node :) 

So it's as you said, the pseudo-code says that we return all the resuming points from the indexer that has the lowest immutable resume point.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: I forgot to commit :)

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: Good point. I'll change that.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: Can you revalidate now?

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
berewt: Don't we need to intersect it with other indexers? Hmm... maybe not, we'll see when we'll implement it. Though, it should be `minimum`, not `maximum`, right?

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
berewt: Yes, all good.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: Mmm.. But at each event, we still need to update the database right? So querying won't add too much overhead, afaik? The main issue I see is that it complexifies the implementation because we need to keep track of the last sync points by only keeping K points, and removing old points after each update.

berewt: No, we don't. The list of points we consider for resuming is _almost_ immutable: as long as we drain, the database shouldn't change.
I say _almost_ because we may want to remove points from this list when we catch up, to ease the processing.

koslambrou: I'll revisit.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
```diff
@@ -45,16 +45,89 @@ Decision
...
+      head $ sortOn (\rps -> (maximum $ filter isImmutablePoint rps, maximum rps)) resumingPoints
+
+  Next, we need a function which processes each ``ChainSyncEvent`` returned by the chain-sync client for each indexer.
+  The pseudo-code is as follows:
+
+  .. code-block:: haskell
+
+    -- Preconditions:
+    --   * The 'initialLastSyncPoints' param should be sorted in ascending order on SlotNo
+    --   * Each element of the 'initialLastSyncPoints' param should satisfy the property 'getSlotNo point == getSlotNo (point + 1) + 1'.
```
berewt, line 73: It sounds unclear to me (and probably wrong, as it implies `getSlotNo point - 1 = getSlotNo (point + 1)` and thus `slotNo`s decrease when we move to a next point)

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
```diff
@@ -45,16 +45,89 @@ Decision
...
+    applyChainSyncEvent initialLastSyncPoints previousChainSyncEventPoint event indexer = do
+      case event of
+        Resume point -> do
```
berewt, line 76: What is a `Resume` event? Is it the initial rollback?

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
```diff
@@ -45,16 +45,89 @@ Decision
...
+            Just _ ->
+              update initialLastSyncPoints $ filter (\p -> getSlotNo p > getSlotNo point)
+        RollBackward point -> do
+          rewind point indexer
+        RollForward block point -> do
+          case initialLastSyncPoints of
+            -- Processed all initial last sync points. The indexer is in sync.
+            -- So we just index the block.
+            [] -> do
+              insert block indexer
```
berewt, line 90: Implementation details: we can probably check if the tail of `initialLastSyncPoints` is empty at the end of a `RollForward` event rather than waiting on the next call.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
```diff
@@ -45,16 +45,89 @@ Decision
...
+            -- We rewind to the previous point, insert the block and remove all points from the
+            -- initial last sync point list.
+            -- We drain the event and remove the last sync point from the initial last sync points.
+            (lowestLastSyncPoint:_) | getSlotNo point == getSlotNo lowestLastSyncPoint -> do
+                rewind previousChainSyncEventPoint indexer
+                insert block indexer
+                update initialLastSyncPoints $ \_ -> []
+
+            -- The chain sync event slot number is higher than the lowest last sync point of the
+            -- indexer. Should not happen though.
```
berewt, line 108: Do all the blocks contain the same amount of slots I thought so but IIRC. it doesn't look alike in cardano-scan.

berewt: If they are different though, it probably means that a we need to rollback to the previous block, so I would suggest a rollback to the previous block too in this case.

berewt (approved):
So minor comments but it's quite close to what I had in mind. If you're convinced too, it should be fine.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: Did you see the `+1` after `(point + 1)`?

berewt: yes and I computed `getSlotNo point == getSlotNo (point + 1) + 1  ~~> getSlotNo point - 1 == getSlotNo (point + 1)`

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: That's right. Initially we didn't need it, but I think we need it right now because of the initial resume points we pass to the chain sync client.

Ex. let's say we have the following resuming points for indexers A and B
A: [1,2,3,4,5]
B: [4,5,6]

Given current resuming point selection, we will use `[1,2,3,4,5]` as the resuming points, and the chain-sync client will probably select the point 5.

So, given current implementation, the first event will be a `RollBackward`. However, we don't want B to rewind to point 5, we want to _ignore_ this RollBackward. However, how do we tell B to ignore this initial RollBackward? My solution was to introduce a Resume event.

Do you see a better way?

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: I don't see how you want to write this though

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: By "Do all the blocks contain the same amount of slots", do you mean if each slot produces a block? Then the answer is no.

Also, are you talking about this line specifically?
```
rewind previousChainSyncEventPoint indexer
```

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
berewt: Not at the moment, I don't like to add one type of event that we need to carry along during all the process while it can happen only at startup. Maybe handling the first Rollback in a different function? (`resume` would call `initiateChainSync` that would deal with the initial rollback before passing it to `applyChainSyncEvent`?)

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
berewt: Oh right, there's no recursive call in `applyChainSyncEvent` nevermind then, the logic is correct, that's the most important part.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
berewt: No I meant "Do each block contain the same amount of slots" if it not the case and if a rollback to block `N` happen, the new block `N+1` may have a different size than the previous block `N+1` and thus the slot number wont match.

berewt: And I was proposing to add a `rollback` before the insert in the lines below this comment (L110).

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: In the end, that type will be conformant with consensus. They also have a different type for resuming and for rollbacks. It's just that the dev that worked on cardano-streaming decided to create an event that combines the two, but the decision was arbitrary afaik.

koslambrou: > Maybe handling the first Rollback in a different function? (resume would call initiateChainSync that would deal with the initial rollback before passing it to applyChainSyncEvent?)

The problem is that you don't know which point the local node will use to initiate the chain sync protocol, so that won't work.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: I still don't understand what you mean. Blocks don't contain slots. Blocks are created a specific slot.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
berewt: We still know that the first event is a rollback. Although, if Consensus already expose a specific event for resuming, we can use it as well.
In that case though, the remaining question (outside of the scope of this ADR) is probably what will we do with the `Resume` event in the general case (we may throw an error for an unexpected event or restart the resuming protocol though.
Ok, let's go for it.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
berewt: Sorry, I wasn't specific enough: I was talking about the distance (in slots) between two blocks.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: Reading back on this, your solution works as well. IIUC, you want to pass a callback function as a parameter to the `withChainSyncStreaming` function which is called on the first resume event, instead of handling a `Resume` event for every worker.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: Ok, but the distance between slots seems irrelevant to me in this situation.

The only situation where we get in this case statement is if we started the chain-sync client with a resuming point that is higher than this indexer's resuming point. If that happens, then there's a bug in the implementation. 

For example, if we get a `RollForward` event in point `10`, but the lowest resuming point of the indexer is `5`, then the initial points we provided to the chain-sync client are wrong. I guess I should probably just do an `error "..."` here?

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
berewt: yes, but it's implementation details, and can maybe be omitted in the adr?

berewt: so we can go with one or the other, as long as it is correct.

koslambrou: Yep let's omit it for the ADR. The principle is the same.

doc/read-the-docs-site/adr/0006-indexer-resuming-strategy.rst
koslambrou: You're totally right :+1: 

I should be defined in terms of blockNo then. `getBlockNo point - 1 == getBlockNo (succ point)`

berewt: Now I get it :)

PR #125, opened 2023-08-05
PLT-6892 Add node tip info to the getLastSyncedBlock query

berewt:
Added the chain tip to the event and propagated it down to the side chain query.

The only issue (that would be complex to overcome) is that we ignore chainTip on rollback.
Handling it means either adding an indexing call on rollback and changing the event type to allow it, or adding a specific backdoor function to change the content of the tip. 

These solutions look ugly so I didn't integrate it. May be there is a cleaner one but I hadn't find it.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
@@ -79,7 +79,7 @@ import Data.List (sortOn)
 import Data.List.NonEmpty (NonEmpty)
 import Data.Map (Map)
 import Data.Map qualified as Map
-import Data.Maybe (catMaybes, listToMaybe, mapMaybe)
+import Data.Maybe (catMaybes, fromMaybe, listToMaybe, mapMaybe)
 import Data.Ord (Down (Down), comparing)
 import Data.Text qualified as Text
 import Data.Text.Encoding qualified as Text
@@ -414,7 +414,7 @@ instance ToJSON UtxoRow where
 
 data instance StorableResult UtxoHandle
   = -- | Result of a 'QueryUtxoByAddress' query
-    UtxoByAddressResult {getUtxoByAddressResult :: ![UtxoResult]}
+    UtxoByAddressResult {getUtxoByAddressResult :: !([UtxoResult], C.ChainTip)}
   | -- | Result of a 'LastSyncedBlockInfoQuery'
     LastSyncedBlockInfoResult {getLastSyncedBlockInfo :: !(WithOrigin BlockInfo)}
   deriving (Eq, Show, Ord)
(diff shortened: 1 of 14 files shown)


marconi-sidechain/src/Marconi/Sidechain/Api/Routes.hs
```diff
@@ -102,21 +102,44 @@ type RpcEpochNonceMethod =
...
           (At (BlockInfo sn bhh bn bt en)) ->
             [ "blockNo" .= bn
             , "blockTimestamp" .= bt
             , "blockHeaderHash" .= bhh
             , "slotNo" .= sn
             , "epochNo" .= en
+            , "nodeTip" .= toJSON tip
             ]
-          Origin -> []
+          Origin -> ["nodeTip" .= toJSON tip]
```
koslambrou, line 142: For consistency, how about *not* setting the "nodeTip" field like the rest of the fields in this scenario? It will just be null anyway.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -1026,33 +1041,40 @@ instance Queryable UtxoHandle where
              GROUP BY s.slotNo
              ORDER BY s.slotNo DESC
              LIMIT ?|]
+        queryTip =
```
koslambrou, line 1044: Should we remove the old chain tips and only keep the most recent one?

Also, I don't see the insert statement of the chainTip. Did I miss it?

berewt: Yes, I may have fat findgered a late rebase, it was there and it isn't anymore, let me fix this before you continue.

marconi-chain-index/src/Marconi/ChainIndex/Indexers/Utxo.hs
```diff
@@ -742,6 +753,16 @@ instance Buffered UtxoHandle where
                    )
                    VALUES (?, ?)|]
             datumRows
+          `concurrently_` do
+            SQL.execute
```
koslambrou, line 757: Do we need a delete statement to remove old chain tips? We only need the most recent one.

berewt: I was sure it was added, adding it again, for real this time.

berewt: I decided to use a constraint to ensure that there's only one row and used `INSERT OR REPLACE`, it's safer and cleaner than `DELETE/INSERT`.

koslambrou: That works

koslambrou (approved):

PR #147, opened 2023-09-01
PLT-634 Trace CLI args

neilmayhew:
This implements:

* Tracing the `CliArgs` on startup
* Add `Trace` and `CliArgs` to `SidechainEnv`
* Add basic tests for `SidechainEnv`
* Add some use of `ReaderT`
* Add a `/params` API endpoint

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-sidechain/doc/API.adoc
+++ b/marconi-sidechain/doc/API.adoc
@@ -1,14 +1,33 @@
-= JSON-RPC API
+= HTTP API
 :toc: left
 :toclevels: 2
 
-This document describes the JSON-RPC HTTP API that end-users will interact with in order to query the data indexed by `marconi-sidechain`.
+This document describes the HTTP API that end-users interact with in order to query the data indexed by `marconi-sidechain`.
 
-== HTTP endpoint
+== HTTP endpoints
 
-In this application, there is a single HTTP endpoint/route: `POST /json-rpc`.
-This endpoint will listen and respond to well-formed link:https://www.jsonrpc.org/specification[JSON-RPC version 2] requests and responses.
-Each of the following subsections reflect the JSON-RPC methods that can be used in the JSON-RPC request.
+In this application, the following HTTP endpoints/routes are available:
+
+[%header%autowidth,format=csv]
+|===
+Route,Methods,ContentType
+`/json-rpc`,`POST`,JSON
+`/addresses`,`GET`,Text
+`/metrics`,`GET`,Text
+`/params`,`GET`,JSON
+`/time`,`GET`,Text
+|===
+
+The `/json-rpc` endpoint responds to well-formed link:https://www.jsonrpc.org/specification[JSON-RPC version 2] requests.
+The JSON-RPC methods that can be used are described <<_supported_json_rpc_methods,below>>.
+
+The `/addresses` endpoint returns the list of target addresses represented as Bech32 text in a JSON string array.
+
+The `/metrics` endpoint returns Prometheus metrics for monitoring the runtime behavior.
+
+The `/params` endpoint returns the parsed command-line parameters as a JSON object.
+
+The `/time` endpoint returns the current time as text. It's typically used for testing the http server outside of JSON-RPC.
 
 == JSON-RPC HTTP endpoint request and response format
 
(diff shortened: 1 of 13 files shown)


neilmayhew:
This is ready for review now

marconi-sidechain/src/Marconi/Sidechain/CLI.hs
```diff
@@ -39,11 +47,21 @@ data CliArgs = CliArgs
...
   }
-  deriving (Show)
+  deriving (Show, Generic, FromJSON, ToJSON)
+
+-- Add orphan instances that aren't provided by Cardano.Api
+deriving instance Generic C.NetworkId
+instance FromJSON C.NetworkMagic
+instance FromJSON C.NetworkId
+instance ToJSON C.NetworkMagic
+instance ToJSON C.NetworkId
```
berewt, line 57: We have a module where we gather all the orphans of Cardano.Api we need, to avoid duplication across packages (`Marconi.ChainIndex.Orphans`), moving these instances here would be great.

marconi-sidechain/src/Marconi/Sidechain/Run.hs
```diff
@@ -35,10 +38,12 @@ run = do
...
     securityParam <- withNodeConnectRetry trace optionsRetryConfig socketFilePath $ do
       Utils.toException $ Utils.querySecurityParam @Void networkId socketFilePath
 
-    rpcEnv <- mkSidechainEnvFromCliArgs securityParam cliArgs
+    rpcEnv <- mkSidechainEnvFromCliArgs securityParam cliArgs trace
     race_
-      (runHttpServer rpcEnv) -- Start HTTP server
-      (runSidechainIndexers trace cliArgs rpcEnv) -- Start the Sidechain indexers
+      (runReaderT runHttpServer rpcEnv) -- Start HTTP server
+      (runReaderT runSidechainIndexers rpcEnv) -- Start the Sidechain indexers
```
berewt, line 49: I wonder if we shouldn't go with `ReaderT` in `runHttpServer` and `runSidechainIndexers` too (all the way down to where we're really using the reader context), but we could keep it for later refactoring.

berewt (approved):
Great work. One minor refactoring, and an item that we can either try to integrate to the PR or save for the future (it's up to you).

marconi-sidechain/doc/API.adoc
```diff
@@ -1,14 +1,33 @@
...
+`/params`,`GET`,JSON
+`/time`,`GET`,Text
+|===
+
+The `/json-rpc` endpoint responds to well-formed link:https://www.jsonrpc.org/specification[JSON-RPC version 2] requests.
+The JSON-RPC methods that can be used are described <<_supported_json_rpc_methods,below>>.
+
+The `/addresses` endpoint returns the list of target addresses represented as Bech32 text in a JSON string array.
+
+The `/metrics` endpoint returns metrics information as text.
```
koslambrou, line 26:
```suggestion
The `/metrics` endpoint returns Prometheus metrics for monitoring the runtime behavior.
```

marconi-sidechain/src/Marconi/Sidechain/CLI.hs
```diff
@@ -1,17 +1,25 @@
...
 import Data.List.NonEmpty (NonEmpty)
+import GHC.Generics (Generic)
 import Marconi.ChainIndex.CLI qualified as Cli
-import Marconi.ChainIndex.Node.Client.Retry (RetryConfig)
+import Marconi.ChainIndex.Node.Client.Retry (RetryConfig (..))
 import Marconi.ChainIndex.Types (
-  IndexingDepth,
-  ShouldFailIfResync,
+  IndexingDepth (..),
+  ShouldFailIfResync (..),
```
koslambrou, line 22: These (..) aren't needed I think. Weird that you didn't get any warning for those as we're using `Wmissing-import-lists`.

marconi-sidechain/src/Marconi/Sidechain/Run.hs
koslambrou: I agree. Let's do it now as we're already touching this part.

koslambrou (commented):
Additionally, can you print the CLI args at startup on the terminal? That was the original request by the Sidechain team.

neilmayhew:
> Additionally, can you print the CLI args at startup on the terminal? That was the original request by the Sidechain team.

I'm doing that [here] (https://github.com/input-output-hk/marconi/pull/147/files#diff-a48ccffbf3c37ba034c160786cad9fcf31063805aaf9ce572ce475a6bee032cfR41).

marconi-sidechain/src/Marconi/Sidechain/CLI.hs
neilmayhew: The constructors are being used, but in a non-obvious way: for deriving `Generic`

marconi-sidechain/src/Marconi/Sidechain/CLI.hs
neilmayhew: Thanks! I'll definitely move them there. I was already a little uncomfortable about putting them here in `CLI` anyway.

One thing I think I'll add is to use deriving strategies to be a little more specific about how the instance is being defined. This in turn allows me to use a `newtype` deriving strategy for `NetworkMagic` so that it doesn't have a redundant `unNetworkMagic` in the serialization. In other words, we would have

```json
{
  "networkId": {
    "contents": 1,
    "tag": "Testnet"
  }
}
```

instead of

```json
{
  "networkId": {
    "contents": {
      "unNetworkMagic": 1
    },
    "tag": "Testnet"
  }
}
```

marconi-sidechain/src/Marconi/Sidechain/Run.hs
neilmayhew: As we discussed on the call, the changes for this could be somewhat complex. I'd like to get this ticket closed out, since it has already grown in scope a couple of times, so let's create a new ticket for continuing the `ReaderT` refactoring and I'll start working on it immediately.

marconi-sidechain/doc/API.adoc
neilmayhew: Thanks!

marconi-sidechain/src/Marconi/Sidechain/CLI.hs
neilmayhew: If I move the derivations as Nicolas suggested, these will go away.

marconi-sidechain/src/Marconi/Sidechain/Run.hs
neilmayhew: https://input-output.atlassian.net/browse/PLT-7446

PR #148, opened 2023-09-06
PLT-7351 Added resuming support in marconi-chain-index with the new marconi API

koslambrou:
* Added the WithResume transformer to the StandardIndexer stack

* Added resuming in runIndexer of marconi-chain-index

* Added possiblity to resume from a preferred point using the CLI

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers.hs
@@ -7,7 +7,6 @@ module Marconi.ChainIndex.Experimental.Indexers where
 
 import Cardano.Api qualified as C
 import Cardano.BM.Tracing qualified as BM
-import Control.Concurrent (MVar)
 import Control.Monad.Cont (MonadIO)
 import Control.Monad.Except (ExceptT, MonadError, MonadTrans (lift))
 import Data.List.NonEmpty qualified as NonEmpty
@@ -21,8 +20,11 @@ import Marconi.ChainIndex.Experimental.Indexers.MintTokenEvent qualified as Mint
 import Marconi.ChainIndex.Experimental.Indexers.Spent qualified as Spent
 import Marconi.ChainIndex.Experimental.Indexers.Utxo qualified as Utxo
 import Marconi.ChainIndex.Experimental.Indexers.UtxoQuery qualified as UtxoQuery
-import Marconi.ChainIndex.Experimental.Indexers.Worker (StandardWorkerConfig (StandardWorkerConfig))
-import Marconi.ChainIndex.Types (BlockEvent (BlockEvent), TxIndexInBlock)
+import Marconi.ChainIndex.Experimental.Indexers.Worker (
+  StandardWorker (StandardWorker),
+  StandardWorkerConfig (StandardWorkerConfig),
+ )
+import Marconi.ChainIndex.Types (BlockEvent (BlockEvent), SecurityParam, TxIndexInBlock)
 import Marconi.Core.Experiment qualified as Core
 import System.FilePath ((</>))
 
@@ -34,7 +36,8 @@ type instance Core.Point [AnyTxBody] = C.ChainPoint
 and expose a single coordinator to operate them
 -}
 buildIndexers
-  :: Core.CatchupConfig
+  :: SecurityParam
+  -> Core.CatchupConfig
   -> Utxo.UtxoIndexerConfig
   -> MintTokenEvent.MintTokenEventConfig
   -> BM.Trace IO Text
@@ -43,21 +46,26 @@ buildIndexers
       Core.IndexerError
       IO
       -- Query part (should probably be wrapped in a more complex object later)
-      ( UtxoQuery.UtxoQueryIndexer IO
+      ( [C.ChainPoint]
+      , UtxoQuery.UtxoQueryIndexer IO
       , -- Indexing part
         Core.WithTrace IO Core.Coordinator (WithDistance BlockEvent)
       )
-buildIndexers catchupConfig utxoConfig mintEventConfig logger path = do
+buildIndexers securityParam catchupConfig utxoConfig mintEventConfig logger path = do
   let mainLogger = BM.contramap (fmap (fmap $ Text.pack . show)) logger
       txBodyCoordinatorLogger = BM.appendName "txBody" mainLogger
 
-  (blockInfoMVar, blockInfoWorker) <- blockInfoBuilder catchupConfig mainLogger path
+  StandardWorker blockInfoMVar blockInfoWorker <-
+    blockInfoBuilder securityParam catchupConfig mainLogger path
 
-  (utxoMVar, utxoWorker) <- utxoBuilder catchupConfig utxoConfig txBodyCoordinatorLogger path
-  (spentMVar, spentWorker) <- spentBuilder catchupConfig txBodyCoordinatorLogger path
-  (datumMVar, datumWorker) <- datumBuilder catchupConfig txBodyCoordinatorLogger path
-  (_mintTokenMVar, mintTokenWorker) <-
-    mintBuilder catchupConfig mintEventConfig txBodyCoordinatorLogger path
+  StandardWorker utxoMVar utxoWorker <-
+    utxoBuilder securityParam catchupConfig utxoConfig txBodyCoordinatorLogger path
+  StandardWorker spentMVar spentWorker <-
+    spentBuilder securityParam catchupConfig txBodyCoordinatorLogger path
+  StandardWorker datumMVar datumWorker <-
+    datumBuilder securityParam catchupConfig txBodyCoordinatorLogger path
+  StandardWorker _mintTokenMVar mintTokenWorker <-
+    mintBuilder securityParam catchupConfig mintEventConfig txBodyCoordinatorLogger path
 
   let getTxBody :: (C.IsCardanoEra era) => TxIndexInBlock -> C.Tx era -> AnyTxBody
       getTxBody ix tx = AnyTxBody ix (C.getTxBody tx)
@@ -78,7 +86,11 @@ buildIndexers catchupConfig utxoConfig mintEventConfig logger path = do
   mainCoordinator <-
     lift $ standardCoordinator mainLogger [blockInfoWorker, coordinatorTxBodyWorkers]
 
-  pure (queryIndexer, mainCoordinator)
+  resumePoints <- Core.lastSyncPoints (fromIntegral securityParam + 1) mainCoordinator
+
+  -- TODO Create a dedicated return type for it instead of a tuple.
+  -- However, we should wait until we have more stuff in the query side before we do it.
+  pure (resumePoints, queryIndexer, mainCoordinator)
 
 -- | Build and start a coordinator of a bunch of workers that takes an @AnyTxBody@ as an input
 buildTxBodyCoordinator
@@ -88,24 +100,23 @@ buildTxBodyCoordinator
   -> [Core.Worker event (Core.Point event)]
   -> m (Core.Worker (WithDistance input) (Core.Point event))
 buildTxBodyCoordinator logger extract workers =
-  snd <$> coordinatorWorker "TxBody coordinator" logger extract workers
+  Core.worker <$> coordinatorWorker "TxBody coordinator" logger extract workers
 
 -- | Configure and start the @BlockInfo@ indexer
 blockInfoBuilder
   :: (MonadIO n, MonadError Core.IndexerError n, MonadIO m)
-  => Core.CatchupConfig
+  => SecurityParam
+  -> Core.CatchupConfig
   -> BM.Trace m (Core.IndexerEvent C.ChainPoint)
   -> FilePath
-  -> n
-      ( MVar (Block.StandardBlockInfoIndexer m)
-      , Core.WorkerM m (WithDistance BlockEvent) (Core.Point Block.BlockInfo)
-      )
-blockInfoBuilder catchupConfig logger path =
+  -> n (StandardWorker m BlockEvent Block.BlockInfo Core.SQLiteIndexer)
+blockInfoBuilder securityParam catchupConfig logger path =
   let extractBlockInfo :: BlockEvent -> Block.BlockInfo
       extractBlockInfo (BlockEvent (C.BlockInMode b _) eno t) = Block.fromBlockEratoBlockInfo b eno t
       blockInfoWorkerConfig =
         StandardWorkerConfig
           "BlockInfo"
+          securityParam
           catchupConfig
           (pure . Just . extractBlockInfo)
           (BM.appendName "blockInfo" logger)
@@ -114,20 +125,19 @@ blockInfoBuilder catchupConfig logger path =
 -- | Configure and start the @Utxo@ indexer
 utxoBuilder
   :: (MonadIO n, MonadError Core.IndexerError n, MonadIO m)
-  => Core.CatchupConfig
+  => SecurityParam
+  -> Core.CatchupConfig
   -> Utxo.UtxoIndexerConfig
   -> BM.Trace m (Core.IndexerEvent C.ChainPoint)
   -> FilePath
-  -> n
-      ( MVar (Utxo.StandardUtxoIndexer m)
-      , Core.WorkerM m (WithDistance [AnyTxBody]) (Core.Point Utxo.UtxoEvent)
-      )
-utxoBuilder catchupConfig utxoConfig logger path =
+  -> n (StandardWorker m [AnyTxBody] Utxo.UtxoEvent Core.SQLiteIndexer)
+utxoBuilder securityParam catchupConfig utxoConfig logger path =
   let extractUtxos :: AnyTxBody -> [Utxo.Utxo]
       extractUtxos (AnyTxBody indexInBlock txb) = Utxo.getUtxosFromTxBody indexInBlock txb
       utxoWorkerConfig =
         StandardWorkerConfig
           "Utxo"
+          securityParam
           catchupConfig
           (pure . NonEmpty.nonEmpty . (>>= extractUtxos))
           (BM.appendName "utxo" logger)
@@ -136,19 +146,18 @@ utxoBuilder catchupConfig utxoConfig logger path =
 -- | Configure and start the @SpentInfo@ indexer
 spentBuilder
   :: (MonadIO n, MonadError Core.IndexerError n, MonadIO m)
-  => Core.CatchupConfig
+  => SecurityParam
+  -> Core.CatchupConfig
   -> BM.Trace m (Core.IndexerEvent C.ChainPoint)
   -> FilePath
-  -> n
-      ( MVar (Spent.StandardSpentIndexer m)
-      , Core.WorkerM m (WithDistance [AnyTxBody]) (Core.Point Spent.SpentInfoEvent)
-      )
-spentBuilder catchupConfig logger path =
+  -> n (StandardWorker m [AnyTxBody] Spent.SpentInfoEvent Core.SQLiteIndexer)
+spentBuilder securityParam catchupConfig logger path =
   let extractSpent :: AnyTxBody -> [Spent.SpentInfo]
       extractSpent (AnyTxBody _ txb) = Spent.getInputs txb
       spentWorkerConfig =
         StandardWorkerConfig
           "spent"
+          securityParam
           catchupConfig
           (pure . NonEmpty.nonEmpty . (>>= extractSpent))
           (BM.appendName "spent" logger)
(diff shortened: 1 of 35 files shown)


marconi-chain-index/src/Marconi/ChainIndex/Experimental/Runner.hs
```diff
@@ -44,47 +59,77 @@ If you want to start several indexers, use @runIndexers@.
...
+            PP.renderStrict $
+              PP.layoutPretty PP.defaultLayoutOptions $
+                PP.pretty NoIntersectionFoundLog
+    Async.concurrently_
+      (void $ runChainSyncStream `catch` whenNoIntersectionFound)
+      (Core.processQueue eventQueue cBox)
 
--- | Run several indexers under a unique coordinator
+{- | Run several indexers under a unique coordinator
+TODO Remove? The user can just call `mkCoordinator` himself.
```
koslambrou, line 108: @berewt What do you think?

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Worker.hs
```diff
@@ -63,7 +84,29 @@ catchupWorkerWithFilter config eventFilter indexer =
   let chainPointDistance :: Core.Point (WithDistance a) -> WithDistance a -> Word64
       chainPointDistance _ = Distance.chainDistance
       mapEventUnderDistance = fmap sequence . traverse (eventExtractor config)
-   in Core.createWorker (workerName config) mapEventUnderDistance $
-        Core.withTrace (logger config) $
-          Core.withCatchup chainPointDistance (catchupConfig config) $
-            Core.withTransform id (eventFilter . Distance.getEvent) indexer
+   in do
+        let getResumablePoints _securityParam ix = do
+              fmap List.singleton $ Core.lastSyncPoint ix
```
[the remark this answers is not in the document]
berewt: I initially thought of using a typeclass (`HasResumePoints`, in `WithResume`) and I'm not a huge fan of this design. We could also add a different query to get the `n` last points for any n, and we could remove `HasResumePoints` (so we get rid of an ad-hoc typeclass).

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Runner.hs
berewt: I initially introduced this function for consistency with the previous API, I like the idea of being able to start a bunch of workers without having to deal with the coordinator but it's probably to naive and impossible to combine easily with the resuming, so it's probably better to drop it.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Worker.hs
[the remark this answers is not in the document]
koslambrou: By different query, you mean something like `lastSyncPoints :: m [Point event]`?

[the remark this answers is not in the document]
berewt: No I was thinking about reusing the `Queryable` interface. I did find an advantage to `Queryable actually: transformers already know how to handle them (otherwise you need to add extra logic for your code to go through the transformers layers).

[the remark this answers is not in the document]
koslambrou: So create a new datatype query for Queryable for getting the last sync points? Why not just extend `IsSync`?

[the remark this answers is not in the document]
berewt: Because being able to provide access to the `n` last events is a much stronger requirements than knowing the last sync points.
Later, we may want to add ephemeral indexers, these indexers may want to expose their last sync points but may have no interest in providing a list of their last sync points.

berewt (commented):
I wrongly start a review and I'm too lazy to copy paste these comments outside of a review, so here's a list of early feedback. :)

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers.hs
```diff
@@ -78,7 +87,16 @@ buildIndexers catchupConfig utxoConfig mintEventConfig logger path = do
...
 
-  pure (queryIndexer, mainCoordinator)
+  let resumePoints =
+        Core.commonResumePoints
+          [ blockInfoResumePoints
+          , utxoResumePoints
+          , spentResumePoints
+          , datumResumePoints
+          , mintTokenResumePoints
+          ]
```
koslambrou, line 97: Part I don't like. Alternatively, I can just call `Core.lastSyncPoints` in `runIndexer` or in `Run` module to get the starting points for the chain-sync protocol. 

Drawback: an SQL query call (might be slow for resuming when database is large enough).

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Worker.hs
```diff
@@ -20,50 +20,84 @@ import Data.Text (Text)
...
   let chainPointDistance :: Core.Point (WithDistance a) -> WithDistance a -> Word64
       chainPointDistance _ = Distance.chainDistance
       mapEventUnderDistance = fmap sequence . traverse (eventExtractor config)
-   in Core.createWorker (workerName config) mapEventUnderDistance $
-        Core.withTrace (logger config) $
-          Core.withCatchup chainPointDistance (catchupConfig config) $
-            Core.withTransform id (eventFilter . Distance.getEvent) indexer
+   in do
+        -- TODO Change withResume so that we can pass in chain points directly instead of using
+        -- function.
```
koslambrou, line 93: What do you think? Should I just do it here?

marconi-core/src/Marconi/Core/Experiment/Class.hs
```diff
@@ -211,14 +212,19 @@ class AppendResult m event query indexer where
...
-  -- | Last sync of the indexer
+-- | We know how far an indexer went in the indexation of events.
+class (Functor m) => IsSync m event indexer where
+  -- | Last sync point of the indexer.
   lastSyncPoint :: indexer event -> m (Point event)
 
+  -- | Last sync points of the indexer. By default, wrap the point from 'lastSyncPoint' into a list.
+  -- If you need more than one element, then you need to reimplement this function.
+  lastSyncPoints :: Word -> indexer event -> m [Point event]
+  lastSyncPoints _ indexer = List.singleton <$> lastSyncPoint indexer
```
koslambrou, line 223: Maybe remove it and require every indexer to implement it?

marconi-core/src/Marconi/Core/Experiment/Query.hs
```diff
@@ -109,3 +116,18 @@ instance
           dbResult <- extractDbResult
           memoryResult <- extractMemoryResult
           pure $ memoryResult <> dbResult
+
+{- | Calculates the common resume points between the resume points of multiple indexers.
+
+TODO Probably not the best place to put this.
+Ideally, it would go in `WithResume`, but that introduces a cyclic dependency when reusing that
+function in `Coordinator` and `SQLiteAggregateQuery` modules.
```
koslambrou, line 124: Thoughts?

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers.hs
berewt: Yes, it was my fear when I thought about this design. The pro part is that we can easily switch to another design for the resumable points, the con part is the verbosity. That's why I liked the idea of having the coordinator implementing a way to get the resumable points from its workers, it provides a default implementation that could save complexity for the end user.
Actually, just using the resume points from the coordinator should work, no? (if it doesn't we could change the coordinator lastSyncPoints implementation to make it work).

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
```diff
@@ -25,25 +30,43 @@ run appName = do
       includeScript = True
       socketPath = Cli.optionsSocketPath $ Cli.commonOptions o
       networkId = Cli.optionsNetworkId $ Cli.commonOptions o
+      retryConfig = Cli.optionsRetryConfig $ Cli.commonOptions o
```
berewt, line 33: Is it useful to have it in this PR?

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
```diff
@@ -25,25 +30,43 @@ run appName = do
...
+      preferedStartingPoint = Cli.optionsChainPoint $ Cli.commonOptions o
```
berewt, line 34: Same here, especially as, as we discussed, we don't have a clear semantic for this option. Unless there's a real need to add it now, I would postpone it.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Runner.hs
```diff
@@ -17,24 +16,32 @@ import Cardano.Api.Extended.Streaming (
...
+import Marconi.ChainIndex.Types (SecurityParam)
 import Marconi.Core.Experiment qualified as Core
-import Prettyprinter qualified as Pretty
-import Prettyprinter.Render.Text qualified as Pretty
+import Prettyprinter qualified as PP
+import Prettyprinter.Render.Text qualified as PP
 import Streaming qualified as S
 import Streaming.Prelude qualified as S
 
+data RunIndexerLog
```
berewt, line 36: Add comment before the merge.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Runner.hs
```diff
@@ -43,48 +50,41 @@ given indexer.
...
   -> IO ()
-runIndexer trace retryConfig socketPath networkId _startingPoint indexer = do
+runIndexer trace securityParam retryConfig socketPath networkId startingPoints indexer = do
   withNodeConnectRetry trace retryConfig socketPath $ do
-    securityParam <- Utils.toException $ Utils.querySecurityParam @Void networkId socketPath
+    Trace.logInfo trace $
+      PP.renderStrict $
+        PP.layoutPretty PP.defaultLayoutOptions $
+          PP.pretty $
+            LastSyncPointsLog startingPoints
```
berewt, line 71: extract it in a function to improve readability

marconi-core/src/Marconi/Core/Experiment/Class.hs
berewt: You mean remove the default implementation? If so, yes, as it's easy enough to implement.

marconi-core/src/Marconi/Core/Experiment/Indexer/SQLiteIndexer.hs
```diff
@@ -298,13 +318,24 @@ instance
...
--- SQLite3 returned ErrorBusy while attempting to perform close: unable to close due to unfinalized statements or unfinished backups
--- Should use 'withConnection' instead.
+  lastSyncPoints :: Word -> SQLiteIndexer event -> m [Point event]
+  lastSyncPoints n indexer = do
+    let (GetLastSyncPointsQuery query limitParam) = indexer ^. lastSyncPointsQuery
+    resultE <- runExceptT $ runLastSyncQuery (indexer ^. connection) query limitParam n
+    case resultE of
+      -- TODO Revisit this. Should probably throw error. However, can't just add constraint
+      -- 'MonadError IndexerError event' because then it conflicts with other functions which have
+      -- 'MonadError (QueryError query) event)'
```
berewt, line 335: We should write a helper to transition between the 2 as we needed it in different places (`runExceptT` to get the `QueryError`, `either` on the result `throwError` on a `Left`).

marconi-core/src/Marconi/Core/Experiment/Query.hs
berewt: In the `Class` module as a helper to implement `lastSyncPoints` for any indexers that handle several indexers (`Coordinator`, `SQLiteAggregateQuery`...)?

marconi-core/src/Marconi/Core/Experiment/Worker.hs
```diff
@@ -43,8 +43,14 @@ import Marconi.Core.Experiment.Type (
   point,
  )
 
+-- | Worker which also provides direct access to the indexer hidden inside it.
+data WorkerIndexer m input event indexer = WorkerIndexer
+  { workerIndexerVar :: !(MVar (indexer event))
+  , worker :: WorkerM m input (Point event)
+  }
```
berewt, line 50: 👍

marconi-core/src/Marconi/Core/Experiment/Indexer/SQLiteIndexer.hs
```diff
@@ -86,7 +93,10 @@ data SQLRollbackPlan point = forall a.
   }
 
 -- : A new type to get the last sync point from an indexer, the query should not take any param
```
berewt, line 95: fix the haddock

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/SyncHelper.hs
```diff
@@ -52,20 +51,23 @@ syncInsertPlan :: Core.InsertPointQuery
...
   Core.InsertPointQuery [sql|INSERT INTO sync (slotNo, blockHeaderHash) VALUES (?, ?)|]
 
+-- TODO Revisit comment
+
 -- | Query for the last point of an indexer
-syncLastPointQuery :: Core.GetLastSyncQuery
-syncLastPointQuery = Core.GetLastSyncQuery $ syncHistoryQuery 1
+syncLastPointsQuery :: Core.GetLastSyncPointsQuery
+syncLastPointsQuery = Core.GetLastSyncPointsQuery syncHistoryQuery ":limit"
+
```
berewt, line 59: See my comment about GetLastSyncPointsQuery in the `SQLiteIndexer` the `":limit"` is probably useless here.

berewt (commented):
I think we're close, probably just need to revisit it one more time to streamline the user experience.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
koslambrou: No, but it was pretty trivial to add it as it was a copy-paste from the old indexer. However, it makes more sense to put in a separate PR.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
koslambrou: For this one, it was pretty important. The reason is that we needed a way to start syncing from the tip of the local node, so that we can test what happens when rollbacks occur on mainnet. It was also trivial to implement because it's the same logic as the old indexers. 

Of course, the semantics would be eventually enhanced as we discussed, but the current implementation is necessary for testing purposes.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Runner.hs
koslambrou: I think this should be done in a separate PR. Ideally, we would have a typeclass (`class HasLog`) which does this. Something like what RIO this https://hackage.haskell.org/package/rio-0.0.3.0/docs/RIO-Prelude-Logger.html does.

What do you think?

koslambrou: I think this should be done in a separate PR. Ideally, we would have a typeclass (`class HasLog`) which does this. Something like what RIO this https://hackage.haskell.org/package/rio-0.0.3.0/docs/RIO-Prelude-Logger.html does.

What do you think?

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/SyncHelper.hs
koslambrou: What comment are you referring to?

berewt: I may have forgot to validate it. I'd prefer to have a `newtype GetLastSyncPointsQuery = GetLastSyncPointsQuery (Word -> SQL.Query)` to improve type-safety. I know that SQLite isn't type safe and we can't do much about it, but adding an unsafety layer by relying on parameter names encoded as text makes it worse.
That's also why I prefer the existing implementation of `syncHistoryQuery` (In this module) rather than the one proposed in this PR.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Runner.hs
koslambrou: 1ded0f7b06623d1f61ad98b08f34bc4993f74048

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Runner.hs
berewt: We may talk about two different things here. I commented on the noise introduced to enable the `PrettyPrinter` resolution.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Worker.hs
berewt: now that we have the possibility to get it from `IsSync`, can we even use this directly? (but we can pass the points as well)

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/SyncHelper.hs
koslambrou: Yeah you're right. Changing it.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Runner.hs
koslambrou: We always _had_ to do this when using `logInfo` with `PrettyPrinter`. Check below with `logError`. Or in `Marconi.ChainIndex.Node.Client.Retry`. I agree it adds noise, but I think this should be done in a separate PR in order to do the change everywhere in the codebase.

marconi-core/src/Marconi/Core/Experiment/Indexer/SQLiteIndexer.hs
koslambrou: Need some help on this one. I'll schedule some time with you.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers.hs
```diff
@@ -78,7 +86,9 @@ buildIndexers catchupConfig utxoConfig mintEventConfig logger path = do
   mainCoordinator <-
     lift $ standardCoordinator mainLogger [blockInfoWorker, coordinatorTxBodyWorkers]
 
-  pure (queryIndexer, mainCoordinator)
+  resumePoints <- Core.lastSyncPoints (fromIntegral securityParam + 1) mainCoordinator
+
+  pure (resumePoints, queryIndexer, mainCoordinator)
```
berewt, line 91: Add a todo to create a dedicated return type for it. We probably want to wait until we have more stuff in the query side before we do it (unless you want to create this type now and we'll update it).

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/SyncHelper.hs
```diff
@@ -7,26 +7,28 @@ module Marconi.ChainIndex.Experimental.Indexers.SyncHelper (
...
-import Marconi.ChainIndex.Types (SecurityParam)
 import Marconi.Core.Experiment qualified as Core
 
 -- | A simple table to store the passed chainpoints
 syncTableCreation :: SQL.Query
 syncTableCreation =
   [sql|CREATE TABLE IF NOT EXISTS sync (slotNo INT PRIMARY KEY, blockHeaderHash BLOB)|]
 
-{- | Used to keep only the @securityParam + 1@ most recent rows of the sync table.
+{- | NOTE: NOT AT ALL PERFORMANT. DO NOT USE.
```
berewt, line 29: 👍

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
```diff
@@ -25,25 +33,48 @@ run appName = do
...
-    (Cli.optionsChainPoint $ Cli.commonOptions o)
+    startingPoints
     indexers
+
+getStartingPoints :: C.ChainPoint -> [C.ChainPoint] -> NonEmpty C.ChainPoint
+getStartingPoints preferredStartingPoint indexerLastSyncPoints =
+  case preferredStartingPoint of
+    C.ChainPointAtGenesis ->
+      fromMaybe (NonEmpty.singleton C.ChainPointAtGenesis) $
+        NonEmpty.nonEmpty indexerLastSyncPoints
```
berewt, line 79: Can't we use `C.ChainPointAtGenesis` if it's provided (I don't think we have a CLI option for it but we could add one in the future).

marconi-core/src/Marconi/Core/Experiment/Class.hs
```diff
@@ -211,14 +213,42 @@ class AppendResult m event query indexer where
...
 
+  -- | Last sync points of the indexer. By default, it wraps the point from 'lastSyncPoint' into a
+  -- list. If you need more than one element, then you need to reimplement this function.
+  lastSyncPoints :: Word -> indexer event -> m [Point event]
+
+{- | Calculates the common resume points between the resume points of multiple indexers.
+
+This is a helper function to implement the @lastSyncPoints@ functon of 'IsSync' for any indexers
+that handle several indexers (like 'Coordinator' and 'SQLiteAggregateQuery').
+
```
berewt, line 229: 1. Add some explanation about how we compute the final interval.
2. Maybe `commonResumePoints` isn't a good name, as these points aren't _common_. `computeResumePoints`?

berewt (approved):
nitpicking 👍

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
koslambrou: True. In the current CLI, if we don't provide a `ChainPoint` as a CLI option, we assume its genesis. So yeah, that should be reworked in the future story about improving this piece of the CLI.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers.hs
koslambrou: I think it makes more sense to do it once, like you said, we have more stuff on the query side.

PR #149, opened 2023-09-06
PLT-7446 Finish refactoring marconi-sidechain to use ReaderT SidechainEnv

neilmayhew:
Also tidy up `marconi-sidechain/src/Marconi/Sidechain/Bootstrap.hs` a bit

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-sidechain/src/Marconi/Sidechain/Api/HttpServer.hs
+++ b/marconi-sidechain/src/Marconi/Sidechain/Api/HttpServer.hs
@@ -8,9 +8,10 @@
 module Marconi.Sidechain.Api.HttpServer where
 
 import Cardano.Api ()
-import Control.Lens ((^.))
+import Control.Lens (view, (^.))
+import Control.Monad.Except (ExceptT (ExceptT), runExceptT)
 import Control.Monad.IO.Class (liftIO)
-import Control.Monad.Reader (ReaderT, ask, lift)
+import Control.Monad.Reader (ReaderT, ask, lift, runReaderT)
 import Data.Bifunctor (Bifunctor (bimap), first)
 import Data.ByteString qualified as BS
 import Data.Proxy (Proxy (Proxy))
@@ -56,7 +57,7 @@ import Network.JsonRpc.Types (
 import Network.Wai.Handler.Warp (runSettings)
 import Prometheus qualified as P
 import Servant.API ((:<|>) ((:<|>)))
-import Servant.Server (Application, Handler, Server, serve)
+import Servant.Server (Application, Handler (Handler), ServerError, ServerT, hoistServer, serve)
 
 -- | Bootstraps the HTTP server
 runHttpServer :: ReaderT SidechainEnv IO ()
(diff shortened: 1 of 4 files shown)


marconi-sidechain/src/Marconi/Sidechain/Bootstrap.hs
```diff
@@ -41,27 +54,28 @@ import System.FilePath ((</>))
...
+          . AddressUtxo.updateEnvState (env ^. sidechainAddressUtxoIndexer)
   let epochStateCallback :: State EpochStateHandle -> IO ()
       epochStateCallback =
         atomically
           . EpochState.updateEnvState
-            (env ^. sidechainIndexersEnv . sidechainEpochStateIndexer . epochStateIndexerEnvIndexer)
+            (env ^. sidechainEpochStateIndexer . epochStateIndexerEnvIndexer)
   let mintBurnCallback :: State MintBurnHandle -> IO ()
-      dbPath = CLI.dbDir cliArgs
+      dbPath = dbDir
```
koslambrou, line 71: Remove?

marconi-sidechain/src/Marconi/Sidechain/Bootstrap.hs
```diff
@@ -70,21 +84,22 @@ runSidechainIndexers = do
...
-      (CLI.socketFilePath cliArgs)
-      (CLI.networkId cliArgs)
+      optionsRetryConfig
+      socketFilePath
+      networkId
       C.ChainPointAtGenesis
-      (CLI.minIndexingDepth cliArgs)
-      (CLI.optionsFailsIfResync cliArgs)
+      minIndexingDepth
+      optionsFailsIfResync
```
koslambrou, line 104: Rationale for removing the qualified import?

marconi-sidechain/src/Marconi/Sidechain/Api/HttpServer.hs
```diff
@@ -68,116 +69,101 @@ runHttpServer = do
...
+marconiApp env =
+  serve (Proxy @API) $
+    hoistServer (Proxy @API) (hoistHandler env) httpRpcServer
 
-jsonRpcServer
-  :: SidechainEnv
-  -- ^  Utxo Environment to access Utxo Storage running on the marconi thread
-  -> Server JsonRpcAPI
-jsonRpcServer env =
+type ReaderHandler env = ExceptT ServerError (ReaderT env IO)
```
koslambrou, line 76: Is there some function that's throwing a `ServerError`?

marconi-sidechain/src/Marconi/Sidechain/Bootstrap.hs
neilmayhew: Good call - thanks. Forgot to make that change.

marconi-sidechain/src/Marconi/Sidechain/Bootstrap.hs
neilmayhew: The changes in this file aren't needed for the ticket, but the file was changed for the previous ticket and I thought I'd tidy things up while I was still here. I think it's much more readable with the `RecordWildCards` instead of the qualified import and explicit record accessors, but I realize that may not fit with the coding style in the rest of the project. If you're not in favour, I'll happily remove these changes.

marconi-sidechain/src/Marconi/Sidechain/Api/HttpServer.hs
neilmayhew: Potentially any handler can. The standard way of returning a non-200 status is to use `throwError err400` or something similar.

The type `Handler` is actually a `newtype` of `ExceptT ServerError IO` and the hoisting just converts the `ReaderHandler` type here into that. It's changing the base monad from `IO` to `ReaderT env IO`, but to do that you have to unwrap the `ReaderHandler` down to `IO (Either ServerError a)` and then wrap it back up to `Handler` (ie `ExceptT ServerError IO`).

I haven't made `ReaderHandler` a `newtype` like `Handler` because I don't think that's needed when it's being used only within this file.

neilmayhew: FYI, `servant` doesn't have a good story around "real" (runtime) exceptions. They get converted into generic 500-status responses, which don't fit the schema for JSON-RPC errors (ie they're just text). I have a solution for this too, if you think it's worth pursuing.

berewt (commented):
A few discussion items. None of them are blocking but we may want to consider them.

marconi-sidechain/src/Marconi/Sidechain/Bootstrap.hs
neilmayhew: I see your point. In this case, I felt it was OK because the fields are all named explicitly in the import, so if you want to know where one of the names comes from (as you would even with a local variable) a search will find it immediately in the import. However, that's still not as immediate as seeing the qualification right at the point of use.

I'll go ahead and revert these changes.

marconi-sidechain/src/Marconi/Sidechain/Bootstrap.hs
```diff
@@ -41,50 +54,51 @@ import System.FilePath ((</>))
...
         ,
-          ( mintBurnWorker mintBurnCallback $ CLI.targetAssets cliArgs
-          , Just $ dbPath </> mintBurnDbName
+          ( mintBurnWorker mintBurnCallback targetAssets
+          , Just $ dbDir </> mintBurnDbName
           )
         ]
+
   lift $
     runIndexers
```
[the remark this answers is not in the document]
neilmayhew: I did consider this, but because `runIndexers` is in `marconi-chain-index` and `SidechainEnv` is in `marconi-sidechain` I thought it would break encapsulation.

Also, looking at the code for `runIndexers`, the uses of the parameters are spread throughout the function, so I don't think the use of `ReaderT` with a more specific env record would add much value. Also, I don't think the record would be reusable anywhere else.

However, if you can see advantages to it, I'm happy to give it a go.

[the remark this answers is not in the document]
berewt: Yes, we can't use `SidechainEnv` as it would indeed create a dependency loop. And definitely, it isn't worth the effort if we need to split in the next calls.

berewt (approved):

PR #150, opened 2023-09-07
PLT-6832 Allow params to be empty for some JSON-RPC methods

neilmayhew:
* `getCurrentSyncedBlock` method
* `getTargetAddresses` method
* Update documentation
* Improve error reporting in `marconi-sidechain/examples/run-pre-prod-queries`

Note that we allow `params` to be empty but not missing, because we don't want `params` to be optional for other methods and they all share a common `Request p` datatype.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/json-rpc/src/Network/JsonRpc/Types.hs
+++ b/marconi-chain-index/json-rpc/src/Network/JsonRpc/Types.hs
@@ -24,6 +24,7 @@ module Network.JsonRpc.Types (
   Request (..),
   JsonRpcErr (..),
   JsonRpcResponse (..),
+  UnusedRequestParams (..),
 
   -- ** Smart constructors for standard JSON-RPC errors
   mkJsonRpcParseErr,
@@ -40,7 +41,7 @@ import Control.Applicative (liftA3, (<|>))
 import Data.Aeson (
   FromJSON (parseJSON),
   ToJSON (toJSON),
-  Value (Null),
+  Value (Null, Object, String),
   object,
   withObject,
   (.:),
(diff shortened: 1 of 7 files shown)


marconi-sidechain/doc/API.adoc
```diff
@@ -245,7 +258,7 @@ Taking the chainpoint before ensure that we have consistent infomation across al
 
 ==== Pre-conditions
 
-None as there are no required inputs.
+None as the input parameter is unused.
```
berewt, line 261: Now that I have a closer look at it, isn't both this version and the original one inaccurate? There is indeed a precondition on the parameter, as we don't accept arbitrary values.

marconi-sidechain/src/Marconi/Sidechain/Bootstrap.hs
berewt, line 1: Just be careful with the last merge/rebase as it seems that you used your other branch as a base for this PR.

marconi-chain-index/json-rpc/src/Network/JsonRpc/Types.hs
```diff
@@ -177,6 +178,17 @@ instance (ToJSON e, ToJSON r) => ToJSON (JsonRpcResponse e r) where
...
+
+instance FromJSON UnusedRequestParams where
+  parseJSON Null = pure UnusedRequestParams
+  parseJSON (String "") = pure UnusedRequestParams
+  parseJSON (Object o) | null o = pure UnusedRequestParams
+  parseJSON _ = fail "The param value must be empty (use '{}', 'null' or empty string)"
+
+instance ToJSON UnusedRequestParams where
+  toJSON = const Null
+
```
berewt, line 191: 👍

berewt (approved):
Just a documentation point, if you agree with this one, can you see if we have a similar precondition problem elsewhere?

marconi-sidechain/doc/API.adoc
neilmayhew: Good catch. I'll change it. This applies to both `getCurrentSyncedBlock` and `getTargetAddresses`.

PR #152, opened 2023-09-07
PLT-7086 Improve error reporting for incorrect --match-asset-id values

neilmayhew:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/CLI.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/CLI.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PolyKinds #-}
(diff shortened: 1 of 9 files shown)


neilmayhew:
The goofy error message is ultimately coming from `cardano-api` and should read "Unable" rather than "Enable". This typo occurs in two places. Is it OK if I submit a PR to fix the typo there?

This PR for `marconi` will still be needed, though, because we don't want the error message to mention `ScriptHash`.

koslambrou:
> The goofy error message is ultimately coming from cardano-api and should read "Unable" rather than "Enable". This typo occurs in two places. Is it OK if I submit a PR to fix the typo there?

Yes! Definitely encouraged to do that

marconi-chain-index/src/Marconi/ChainIndex/CLI.hs
```diff
@@ -317,17 +318,28 @@ commonMaybeTargetAssetParser =
...
-      parseAssetName = either (fail . show) pure . C.deserialiseFromRawBytesHex C.AsAssetName . Text.encodeUtf8
+      parseAssetName =
+        either (fail . C.displayError) pure . C.deserialiseFromRawBytesHex C.AsAssetName . Text.encodeUtf8
 
       parsePolicyId :: Text -> Opt.ReadM C.PolicyId
-      parsePolicyId = either (fail . show) pure . C.deserialiseFromRawBytesHex C.AsPolicyId . Text.encodeUtf8
+      parsePolicyId =
+        either (fail . displayError') pure . C.deserialiseFromRawBytesHex C.AsPolicyId . Text.encodeUtf8
+
+      displayError' =
```
koslambrou, line 328: Add a comment on why we need this vs just using `displayError`. You can just say like in your previous comment that we don't want to mention `ScriptHash` in the error message.

koslambrou (approved):
LGTM. Good work :)

PR #163, opened 2023-09-26
PLT-1393 Begin implementing an RPC API for the experimental indexers

neilmayhew:
Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


marconi-chain-index/src/Marconi/ChainIndex/Experimental/Api/HttpServer.hs
```diff
@@ -0,0 +1,339 @@
...
+  getTimeHandler
+    :<|> getParamsHandler
+    :<|> getTargetAddressesHandler
+    :<|> getMetricsHandler
+
+httpRpcServer :: ReaderServer HttpServerConfig API
+httpRpcServer = jsonRpcServer :<|> restApiServer
+
+-- | Convenience function for using @MVar@-based indexers
+viewMVar :: (MonadReader s m, MonadIO m) => Getting (MVar a) s (MVar a) -> m a
```
berewt, line 169: Do you need `MonadReader` here?

neilmayhew: I think so, yes. The [type](https://hackage.haskell.org/package/lens-5.2.3/docs/Control-Lens-Combinators.html#v:view) of `view` is

```haskell
view :: MonadReader s m => Getting a s a -> m a
```

berewt: Oh I didn't notice that `view` relied on the `MonadReader`, my bad.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Api/Query/Indexers/Utxo.hs
berewt, line 1: It is a WIP that mixes the old and new indexers part?

berewt (commented):
The general design of the HTTP server and its integration looks fine.

It's hard to review it at the current stage as it's hard to tell which part of the comment is for future work included in this ticket, which part will be deleted and which part is forthcoming work for other tickets.

As discussed yesterday, I suggest that we move forward as soon as we have one endpoint ready and we'll create stories for the missing endpoints.

If you need the updates on the EpochState, you can rebase on this branch: https://github.com/input-output-hk/marconi/pull/181 (or we merge it and you rebase on main after that, both are okay)

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Api/Query/Indexers/Utxo.hs
neilmayhew: It was. I've deleted those parts.

neilmayhew:
Some of the queries are failing below the RPC level. The output from `marconi-sidechain/examples/run-pre-prod-queries` looks like this:

```
=============================================
== getNonceByEpoch
=============================================

getNonceByEpoch(3) =
{
  "error": {
    "code": -32602,
    "data": "The required point is ahead of the current index",
    "message": "Invalid params"
  },
  "id": 1,
  "jsonrpc": "2.0"
}

getNonceByEpoch(4) =
{
  "error": {
    "code": -32602,
    "data": "The required point is ahead of the current index",
    "message": "Invalid params"
  },
  "id": 1,
  "jsonrpc": "2.0"
}

=============================================
== getActiveStakePoolDelegationByEpoch
=============================================

getActiveStakePoolDelegationByEpoch(7) =
{
  "error": {
    "code": -32602,
    "data": "The required point is ahead of the current index",
    "message": "Invalid params"
  },
  "id": 1,
  "jsonrpc": "2.0"
}

getActiveStakePoolDelegationByEpoch(31) =
{
  "error": {
    "code": -32602,
    "data": "The required point is ahead of the current index",
    "message": "Invalid params"
  },
  "id": 1,
  "jsonrpc": "2.0"
}
```

These RPC errors correspond to an `AheadOfLastSync` `QueryError`.

neilmayhew:
I've rebased this on `main` but I haven't got a fix for the nix evaluation errors yet

marconi-core-json-rpc/src/Marconi/Core/JsonRpc.hs
```diff
@@ -55,14 +148,30 @@ queryIndexerVarHttpHandler indexerVar q = do
...
+      "The requested point is too far in the past and has been pruned"
+  Core.AheadOfLastSync mResult ->
+    mkJsonRpcInvalidParamsErr . Just $
+      "The required point is ahead of the current index"
+        <> case mResult of
+          Nothing -> ""
+          Just r -> ". The latest result was: " <> show r
+  Core.IndexerQueryError msg ->
+    mkJsonRpcInternalErr . Just $
+      "The indexer query failed: " <> Text.unpack msg
```
berewt, line 177: We may like a pretty instance for errors that we would use it, although it can be done in a separate ticket.

berewt (approved):
Looks good. For the comment, it's up to you: we can either address it in your PR or do it in another ticket.

marconi-core-json-rpc/src/Marconi/Core/JsonRpc.hs
neilmayhew: Let's do it in a separate ticket

neilmayhew:
I've added a hack to get this to build in nix. However, we won't want to merge that so I've left this as a draft.

I've also added a PoC of adding sidechain handlers. The concept I chose to model is `Profunctor` but I didn't think it was appropriate to bring in the `profunctors` library as a dependency.

The PoC handler uses some query and result types based loosely on the ones currently in use in `marconi-sidechain`. The details aren't important, however, and I just wanted to show how it could be done. The type of `dimapHandler` is very general so it will work for both types of handler (our `ReaderHandler` and the plain servant `Handler`).

marconi-core-json-rpc/src/Marconi/Core/JsonRpc.hs
```diff
@@ -100,6 +101,24 @@ mkHttpRequestTracer trace =
...
+This is technically a profunctor operation with some extra levels of @Functor@
+on the right: @dimapHandler l ≡ dimap l . fmap . fmap@
+
+However, since we need it only for @(->)@ it's not worth pulling in the `profunctor`
+dependency.
+-}
+dimapHandler
+  :: (Functor m)
+  => (qry' -> qry)
+  -> (res -> res')
```
berewt, line 117: I think there are some case where we may need an effectful result here `res -> m res'` (I'm thinking about Wil's current work to add a stability flag to the result, that requires access to `lastStablePoint`, which is effectful).
We can either provide a monadic and a pure variant or just a effectful version, both sounds okay to me.

This aside, I think the design is great.

neilmayhew: It seems to me that if we need an effectful modification of either the query or the result, we're no longer looking at the simple adaptation of one handler into another. In that situation, I think it would be clearer to write the code more directly. Would you be OK with adding an effectful mapping function later if it turns out to be worthwhile?

neilmayhew:
I've done the following:

* Removed the dummy PoC code that uses `dimapHandler` and instead included a simplified example in the Haddock comment
* Disabled cross-compilation at the top level instead of trying to do it piecemeal on a per-project basis (which I was unable to get working)

I haven't added an effectful version of `dimapHandler` for the reasons I explained in my comment. I'll go ahead and merge this as-is and if we want another mapping function maybe we can add that in a separate PR.

Also, if you're not happy with the complete disabling of cross-compilation maybe we can create a ticket to get it working again. However, it seems that cross-compilation was pretty much disabled anyway, and we're better off without it.

berewt:
> I haven't added an effectful version of dimapHandler for the reasons I explained in my comment. I'll go ahead and merge this as-is and if we want another mapping function maybe we can add that in a separate PR.

I didn't get into the details but I think we'll need it. As you said though, it isn't a problem as we can introduce it by extending `dimapHandler` when needed.

> Also, if you're not happy with the complete disabling of cross-compilation maybe we can create a ticket to get it working again. However, it seems that cross-compilation was pretty much disabled anyway, and we're better off without it.

Windows cross-compilation isn't an issue at the moment at least. We can think about it a lot later and I don't see it as a priority. If I read correctly, the `gitRev` has been re-enabled, and that's by far the most important part. So that's good.

PR #168, opened 2023-09-27
Updated marconi-tutorial with additional features

koslambrou:
* Added resuming capabilities

* Used ReaderT pattern and Env type to share types between indexing thread and query thread. Used an IORef to share the MVar between the query thread and the indexer thread.

* Completed link of JSON-RPC server to indexer thread.

* Started experimenting with typeclass to convert a Queryable to a Servant Handler.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
@@ -75,7 +75,7 @@ run appName = do
         Right result -> pure result
       )
 
-  let startingPoints = getStartingPoints preferredStartingPoint indexerLastStablePoint
+  let startingPoint = getStartingPoint preferredStartingPoint indexerLastStablePoint
 
   logInfo trace $ appName <> "-" <> Text.pack Cli.getVersion
 
@@ -85,13 +85,13 @@ run appName = do
         retryConfig
         securityParam
         networkId
-        startingPoints
+        startingPoint
         socketPath
     )
     indexers
 
-getStartingPoints :: C.ChainPoint -> C.ChainPoint -> C.ChainPoint
-getStartingPoints preferredStartingPoint indexerLastSyncPoint =
+getStartingPoint :: C.ChainPoint -> C.ChainPoint -> C.ChainPoint
+getStartingPoint preferredStartingPoint indexerLastSyncPoint =
   case preferredStartingPoint of
     C.ChainPointAtGenesis -> indexerLastSyncPoint
     nonGenesisPreferedChainPoint -> nonGenesisPreferedChainPoint
--- a/marconi-tutorial/src/Marconi/Tutorial/Env.hs
+++ b/marconi-tutorial/src/Marconi/Tutorial/Env.hs
@@ -1,0 +1,34 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+module Marconi.Tutorial.Env where
+
+import Cardano.BM.Trace (Trace)
+import Control.Lens (Lens', makeLenses)
+import Data.Text (Text)
+import Marconi.ChainIndex.Types (SecurityParam)
+import Marconi.Tutorial.CLI qualified as CLI
+import Marconi.Tutorial.Indexers.AddressCount (AddressCountStandardWorker)
+
+data Env = Env
+  { _envIndexers :: IndexersEnv
+  -- ^ Access the Tutorial indexers for querying purposes.
+  , _envCliArgs :: !CLI.Options
+  , _envStdoutTrace :: !(Trace IO Text)
+  , _securityParam :: !SecurityParam
+  }
+
+-- | Should contain all the indexers required by the tutorial.
+newtype IndexersEnv = IndexersEnv
+  { _addressCountIndexerEnv :: AddressCountIndexerEnv
+  }
+
+newtype AddressCountIndexerEnv = AddressCountIndexerEnv
+  { _addressCountIndexerVar :: AddressCountStandardWorker
+  }
+
+makeLenses ''Env
+makeLenses ''IndexersEnv
+makeLenses ''AddressCountIndexerEnv
+
+addressCountIndexerWorker :: Lens' Env AddressCountStandardWorker
+addressCountIndexerWorker = envIndexers . addressCountIndexerEnv . addressCountIndexerVar
--- a/marconi-tutorial/src/Marconi/Tutorial/HttpServer.hs
+++ b/marconi-tutorial/src/Marconi/Tutorial/HttpServer.hs
@@ -1,32 +1,34 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeOperators #-}
 
 module Marconi.Tutorial.HttpServer where
 
+import Control.Concurrent (MVar, readMVar)
+import Control.Lens (to, view, (^.))
+import Control.Monad.Except (ExceptT, runExceptT)
+import Control.Monad.IO.Class (liftIO)
+import Control.Monad.Reader (ReaderT, ask)
 import Data.Data (Proxy (Proxy))
+import Marconi.ChainIndex.Experimental.Indexers.Worker qualified as Core
+import Marconi.Core.Experiment qualified as Core
+import Marconi.Tutorial.CLI (optionsHttpPort)
+import Marconi.Tutorial.Env (
+  Env,
+  addressCountIndexerWorker,
+  envCliArgs,
+ )
+import Marconi.Tutorial.Indexers.AddressCount (AddressCountQuery)
 import Network.JsonRpc.Server.Types ()
-import Network.JsonRpc.Types (JsonRpc, JsonRpcErr, RawJsonRpc)
+import Network.JsonRpc.Types (JsonRpc, JsonRpcErr, RawJsonRpc, mkJsonRpcInternalErr)
 import Network.Wai.Handler.Warp (defaultSettings, runSettings, setPort)
-import Servant.API ((:>))
+import Servant ((:>))
 import Servant.Server (Application, Handler, Server, serve)
 
-runHttpServer :: Maybe Int -> IO ()
-runHttpServer maybePort = do
-  let httpSettings = maybe defaultSettings (flip setPort defaultSettings) maybePort
-  runSettings
-    httpSettings
-    httpApp
-
-httpApp :: Application
-httpApp = serve (Proxy @API) httpServer
-
-httpServer :: Server API
-httpServer = getAddressCountHttpHandler
-
-getAddressCountHttpHandler :: String -> Handler (Either (JsonRpcErr String) Int)
-getAddressCountHttpHandler _ =
-  pure $ Right 1
+{----------------------
+   API specification
+-----------------------}
 
 type API = "api" :> RawJsonRpc RpcMethod
 
@@ -35,6 +37,62 @@ type RpcMethod = RpcAddressCountMethod
 type RpcAddressCountMethod =
   JsonRpc
     "getAddressCount"
+    AddressCountQuery
     String
-    String
-    Int
+    (Core.Result AddressCountQuery)
+
+{----------------------
+   API implementation
+-----------------------}
+
+runHttpServer :: ReaderT Env IO ()
+runHttpServer = do
+  env <- ask
+  let cliOptions = view envCliArgs env
+  let httpSettings = maybe id setPort (optionsHttpPort cliOptions) defaultSettings
+  liftIO $ runSettings httpSettings (httpApp env)
+
+httpApp
+  :: Env
+  -> Application
+httpApp env = serve (Proxy @API) (httpServer env)
+
+httpServer
+  :: Env
+  -> Server API
+httpServer = getAddressCountHttpHandler
+
+getAddressCountHttpHandler
+  :: Env
+  -> AddressCountQuery
+  -> Handler (Either (JsonRpcErr String) (Core.Result AddressCountQuery))
+getAddressCountHttpHandler env =
+  queryIndexerVarHandler (env ^. addressCountIndexerWorker . to Core.standardWorkerIndexerVar)
+
+queryIndexerVarHandler
+  :: ( Show (Core.Result query)
+     , Ord (Core.Point event)
+     , Core.Queryable (ExceptT (Core.QueryError query) Handler) event query indexer
+     , Core.IsSync (ExceptT (Core.QueryError query) Handler) event indexer
+     )
+  => MVar (indexer event)
+  -> query
+  -> Handler (Either (JsonRpcErr String) (Core.Result query))
+queryIndexerVarHandler indexerVar q = do
+  indexer <- liftIO $ readMVar indexerVar
+  toHttpHandler indexer q
+
+toHttpHandler
+  :: ( Show (Core.Result query)
+     , Ord (Core.Point event)
+     , Core.Queryable (ExceptT (Core.QueryError query) Handler) event query indexer
+     , Core.IsSync (ExceptT (Core.QueryError query) Handler) event indexer
+     )
+  => indexer event
+  -> query
+  -> Handler (Either (JsonRpcErr String) (Core.Result query))
+toHttpHandler indexer q = do
+  countE <- runExceptT $ Core.queryLatest q indexer
+  case countE of
+    Left err -> pure $ Left $ mkJsonRpcInternalErr (Just $ show err)
+    Right count -> pure $ Right count
--- a/marconi-tutorial/src/Marconi/Tutorial/Indexers.hs
+++ b/marconi-tutorial/src/Marconi/Tutorial/Indexers.hs
@@ -1,48 +1,84 @@
-{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE FlexibleContexts #-}
 
 module Marconi.Tutorial.Indexers where
 
-import Cardano.BM.Trace (Trace)
-import Data.Text (Text)
-import Data.Void (Void)
+import Cardano.BM.Trace (logError)
+import Control.Lens (view, (^.))
+import Control.Monad.Except (ExceptT, runExceptT)
+import Control.Monad.IO.Class (MonadIO, liftIO)
+import Control.Monad.Reader (MonadReader (ask), ReaderT)
+import Data.Text qualified as Text
 import Marconi.ChainIndex.CLI qualified as CommonCLI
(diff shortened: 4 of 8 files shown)


koslambrou:
@neilmayhew I know you're working on a related issue. Hopefully, any learnings you made in `marconi-sidechain` could be integrated here as well (and vis-versa). Let me know what you think.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
```diff
@@ -64,7 +64,7 @@ run appName = do
         Right result -> pure result
       )
 
-  let startingPoints = getStartingPoints preferredStartingPoint indexerLastStablePoint
+  let startingPoints = getStartingPoint preferredStartingPoint indexerLastStablePoint
```
berewt, line 67: `startingPoints -> startingPoint`

marconi-tutorial/src/Marconi/Tutorial/Env.hs
```diff
@@ -0,0 +1,62 @@
...
+  , _envStdoutTrace :: !(Trace IO Text)
+  }
+
+-- | Should contain all the indexers required by the tutorial.
+newtype IndexersEnv = IndexersEnv
+  { _addressCountIndexerEnv :: AddressCountIndexerEnv
+  }
+
+newtype AddressCountIndexerEnv = AddressCountIndexerEnv
+  { _addressCountIndexerVar :: IORef (MVar AddressCountIndexer)
```
neilmayhew, line 41: It seems a bit ugly to me to have an `IORef` wrapping an `MVar`, since they're both mutable variables. I can see why you have to do it, because `Indexers.runIndexers` is creating the indexers as well as running them. However, perhaps there's a different approach that would eliminate this requirement as well as bringing other improvements.

One would be to separate the creation and running of the indexers, as we do in `marconi-sidechain` at present. Then the indexers can be placed into the environment when it's created. Combined with this, I think it's helpful to have different environments for the http server and the indexers, because they need to know about different things. For example, there's no need for the indexers to know which port the http server should be listening on. I've used this approach for the new RPC server.

Another approach would be to put the `MVar` directly into the environment, but leave it empty initially. `runIndexers` would then put the newly created indexer into the `MVar`, and then use the `MVar` in the same way that it does at present. In the long term, though, we might want to hide the `MVar` as an internal artifact of the indexer, like we do with the Utxo indexer in the new architecture.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
neilmayhew, line 67: Maybe make the variable name singular too?

marconi-tutorial/src/Marconi/Tutorial/Env.hs
```diff
@@ -0,0 +1,62 @@
...
+  }
```
berewt, line 42: I do understand the idea of providing a taste of how we would create an adequate architecture for a larger set of indexers, but with only one indexer, it introduces a lot of indirections.

marconi-tutorial/src/Marconi/Tutorial/HttpServer.hs
```diff
@@ -35,6 +40,42 @@ type RpcMethod = RpcAddressCountMethod
...
+
+{----------------------
+   API implementation
+-----------------------}
+
+runHttpServer :: ReaderT Env IO ()
+runHttpServer = do
+  env <- ask
+  let cliOptions = view envCliArgs env
+  let httpSettings = maybe defaultSettings (flip setPort defaultSettings) (optionsHttpPort cliOptions)
```
neilmayhew, line 55:
```suggestion
  let httpSettings = maybe id setPort (optionsHttpPort cliOptions) defaultSettings
```

marconi-tutorial/src/Marconi/Tutorial/Env.hs
berewt, line 41: Why an `IORef` of an `MVar` instead of just an `MVar` or just an `IORef` (although, we may have harcoded the `MVar` somewhere in the codebase and moving to an `IORef` might not be suitable.)

marconi-tutorial/src/Marconi/Tutorial/Indexers.hs
```diff
@@ -2,44 +2,77 @@
...
+import System.Exit (exitFailure)
 import System.FilePath ((</>))
 
-runIndexers :: Trace IO Text -> CLI.Options -> IO ()
-runIndexers trace o = do
-  createDirectoryIfMissing True (CLI.optionsDbPath o)
+runIndexers :: ReaderT Env IO ()
+runIndexers = do
+  cliOptions <- asks (view envCliArgs)
+  stdoutTrace <- asks (view envStdoutTrace)
```
neilmayhew, line 31:
```suggestion
  cliOptions <- view envCliArgs
  stdoutTrace <- view envStdoutTrace
```

marconi-tutorial/src/Marconi/Tutorial/Indexers.hs
```diff
@@ -2,44 +2,77 @@
...
+  liftIO $ writeIORef addressCountIndexerVar $ Core.standardWorkerIndexerVar addressCountWorker
+
+  indexer <- liftIO $ Core.mkCoordinator $ fmap Core.standardWorker [addressCountWorker]
+
+  indexerLastSyncPointsE <-
+    runExceptT $ Core.lastSyncPoint indexer
+  indexerLastSyncPoints <- case indexerLastSyncPointsE of
+    Left err -> do
+      liftIO $ logError stdoutTrace $ Text.pack $ show err
+      liftIO exitFailure
```
neilmayhew, line 59:
```suggestion
    Left err -> liftIO $ do
      logError stdoutTrace $ Text.pack $ show err
      exitFailure
```

marconi-tutorial/src/Marconi/Tutorial/HttpServer.hs
```diff
@@ -35,6 +40,42 @@ type RpcMethod = RpcAddressCountMethod
...
+getAddressCountHttpHandler
+  :: Env
+  -> AddressCountQuery
+  -> Handler (Either (JsonRpcErr String) (Core.Result AddressCountQuery))
+getAddressCountHttpHandler env q = do
+  let indexerVar =
+        env
+          ^. envIndexers
+            . addressCountIndexerEnv
+            . addressCountIndexerVar
```
berewt, line 77: can we get an alias for this lens?

marconi-tutorial/src/Marconi/Tutorial/HttpServer.hs
```diff
@@ -35,6 +40,42 @@ type RpcMethod = RpcAddressCountMethod
...
+  liftIO $
+    readIORef indexerVar >>= readMVar >>= \indexer -> do
+      r <- Core.query Core.genesis q indexer
```
berewt, line 80: Promote the use of `queryLatest` in the tutorial maybe?

marconi-tutorial/src/Marconi/Tutorial/Indexers/AddressCount.hs
```diff
@@ -54,36 +59,50 @@ import Marconi.Core.Experiment (
...
+  deriving anyclass (FromJSON)
 
 type instance Result AddressCountQuery = Int
 
 -- Note that for Marconi, it is more optimal to return `Nothing` instead of
 -- `Just $ AddressCountEvent mempty`.
-getEventsFromBlock :: WithDistance BlockEvent -> Maybe AddressCountEvent
-getEventsFromBlock (WithDistance _ (BlockEvent (C.BlockInMode (C.Block _ txs) _) _ _)) =
+getEventsFromBlock :: BlockEvent -> Maybe AddressCountEvent
+getEventsFromBlock (BlockEvent (C.BlockInMode (C.Block _ txs) _) _ _) =
```
berewt, line 105: I'm not a huge fan of this pattern matching, but I can't provide a nice alternative.

berewt (commented):
A first batch of comments.

marconi-tutorial/src/Marconi/Tutorial/Indexers/AddressCount.hs
```diff
@@ -228,3 +244,28 @@ mkAddressCountMixedIndexer'
 mkAddressCountMixedIndexer' dbPath keep flush = do
   sqliteIndexer <- mkAddressCountSqliteIndexer dbPath
   pure $ mkMixedIndexer keep flush sqliteIndexer mkListIndexer
+
+-- TEMPORARY --
+-- Trying out some ideas.
```
neilmayhew, line 249: I'm not sure what I think about exposing internal model types directly in the API. The advantage is that avoids a lot of boilerplate (hand-written `aeson` instances, for example). However, the disadvantage is that it couples the internal and external representations, which could in future constrain the implementation if we want to improve the internals without changing the API.

Another consideration, in the general case at least, is that you can end up accidentally exposing private information when a field is added to the internal type. I think it was Slack that was sending (hashed) copies of passwords between workspaces in the traffic between client and server. Someone I worked with at the time pointed out that this would have been impossible if the application had used separate internal and "edge" types, because no-one is going to put a field called "password" in an edge type.

Since our present customers are all internal, these disadvantages probably don't matter too much at the moment, but presumably the situation could change in the future.

marconi-tutorial/src/Marconi/Tutorial/Indexers/AddressCount.hs
neilmayhew: Perhaps a combination of view patterns and lenses?

marconi-tutorial/src/Marconi/Tutorial/Env.hs
neilmayhew: There's also a potential race condition in that the http server could start to query the `MVar` before the worker has been created and the real `MVar` has been rewritten into the `IORef`. Then, the http thread would hang forever because it's waiting on an `MVar` that will never be filled.

marconi-tutorial/src/Marconi/Tutorial/Env.hs
koslambrou: True. We have 2 options:

1- Keep this package as simple as possible as just used for explaining how to write indexer. Then, add a 2nd package called `marconi-starter` which provides the adequate architecture for starting a Marconi project.
2- Have `marconi-tutorial` be both a tutorial + project scaffold.

marconi-tutorial/src/Marconi/Tutorial/Env.hs
koslambrou: That's because in `Indexers` module, we need to share the indexer's `MVar` with the query thread. The issue is that the indexer thread already creates the MVar, and the way I found to share that `MVar` with the query thread is to use an `IORef` where I can just mutate the `MVar` once made available by the indexer thread.

Haven't thought too deeply about it, but if you had an alternative approach, please tell.

marconi-tutorial/src/Marconi/Tutorial/HttpServer.hs
koslambrou: True, but the user experience is better with `query` right now. With `queryLatest`, I get a `MonadError` that I need to handle.

marconi-tutorial/src/Marconi/Tutorial/Indexers/AddressCount.hs
koslambrou: Maybe extend `cardano-api-extended` (no pun intended) with accessors for the fields of `BlockInMode`?

marconi-tutorial/src/Marconi/Tutorial/Env.hs
koslambrou: Ah, Neil proposed the solution to first create the indexers, and then pass the `MVar` to the two threads. Seems like a good approach.

koslambrou: Good idea! I'll go with this approach.

marconi-tutorial/src/Marconi/Tutorial/Env.hs
berewt: I like the `marconi-starter` idea, as using several indexers may lead to other complications, as the need to use a `Coordinator`.

marconi-tutorial/src/Marconi/Tutorial/Env.hs
berewt: I totally forgot where the http server should be plugged but yes, as we use them in both threads, it makes sense to create the `MVar` before we fork the http server.

marconi-tutorial/src/Marconi/Tutorial/HttpServer.hs
berewt: Indeed, but you don't have this error in the general case at the cost of accepting the risk of asynchronous errors sent by sqlite in the `Queryable` instance.
I know it adds complexity but I think we should promote explicit errors here.

marconi-tutorial/src/Marconi/Tutorial/Indexers/AddressCount.hs
berewt: I'm not a huge fan of patterns (I'm probably too old school to like them) but it's a personal preference, so I wouldn't interfere if you think it's a good idea. Putting it, or some standard accessors or lenses in `cardano-api-extended` sounds like a good idea.

neilmayhew: There are pattern synonyms (defined with the `pattern` keyword) and view patterns (used with `->` to pre-apply a function to a binding. Do you dislike both, @berewt, or just one of them?

berewt: Both, the first because it adds complexity to the syntax for few benefits, the second because it obfuscates the original data structure without giving you any hint to notice it. It may look like a type has several constructors, while you only look at one constructor and some patterns.
(again, it's highly subjective)

neilmayhew: I was thinking of something like this:
```suggestion
getEventsFromBlock (preview blockTxs -> Just txs) =
```
where `blockTxs` is a lens.

neilmayhew: I've just discovered that the `C.Block` constructor is itself a pattern! 😂

```haskell
pattern Block :: BlockHeader -> [Tx era] -> Block era
pattern Block header txs <- (getBlockHeaderAndTxs -> (header, txs))
```

berewt: On your proposal, I would advocate that I don't see a major improvement over `getEventsFromBlock bim = let txs = view blockTxs bim in ...` (but we can easily use one or the other, based on personal taste, which is totally fine).
And yes `C.Block` is a pattern, guess which pattern reminds me that I don't like them... :)

neilmayhew: I was thinking that this was one of several equations for the function, and therefore needs to be a guard (hence `preview` rather than `view`) but I see now that it's the only one. In this case, the `let` binding is probably clearer.

However, I noticed that the function currently uses both `let` and `where`. I think it's better to use one or the other but not both in any single function.

marconi-tutorial/src/Marconi/Tutorial/HttpServer.hs
koslambrou: True. Also, I think I misinterpreted `query`. By using `query genesis`, it should basically return nothing, right? So, using `query` with `getLastSyncPoint` should indeed yield a `MonadError`.

berewt: Yes, it should be this way, although, we don't have any mean to enforce it: an indexer implementation could totally ignore the parameter. If you want to yield the errors, there is some helper in the `SQLiteIndexer` package. Promoting them is probably a good idea.

marconi-tutorial/src/Marconi/Tutorial/Env.hs
```diff
@@ -0,0 +1,39 @@
...
+makeLenses ''IndexersEnv
+makeLenses ''AddressCountIndexerEnv
+
+addressCountIndexerWorker :: Lens' Env AddressCountStandardWorker
+addressCountIndexerWorker = lens get set
+  where
+    get =
+      view (envIndexers . addressCountIndexerEnv . addressCountIndexerVar)
+    set env indexerVar =
+      env & envIndexers . addressCountIndexerEnv . addressCountIndexerVar .~ indexerVar
```
neilmayhew, line 39:
```suggestion
addressCountIndexerWorker = envIndexers . addressCountIndexerEnv . addressCountIndexerVar
```

marconi-tutorial/src/Marconi/Tutorial/HttpServer.hs
```diff
@@ -35,6 +37,62 @@ type RpcMethod = RpcAddressCountMethod
...
+     , Core.IsSync (ExceptT (Core.QueryError query) Handler) event indexer
+     )
+  => indexer event
+  -> query
+  -> Handler (Either (JsonRpcErr String) (Core.Result query))
+toHttpHandler indexer q = do
+  countE <- runExceptT $ Core.queryLatest q indexer
+  case countE of
+    Left err -> pure $ Left $ mkJsonRpcInternalErr (Just $ show err)
+    Right count -> pure $ Right count
```
neilmayhew, line 98: Neat!

neilmayhew: FYI, you could use `first` here:
```
toHttpHandler indexer q =
  first (mkJsonRpcInternalErr . Just . show) <$>
    runExceptT (Core.queryLatest q indexer)
```
but maybe that's too terse.

neilmayhew (approved):

marconi-tutorial/src/Marconi/Tutorial/HttpServer.hs
berewt: Will we move it to the future `marconi-json-rpc` library? It looks like a highly reusable piece of code now.

berewt (approved):

marconi-tutorial/src/Marconi/Tutorial/HttpServer.hs
koslambrou: > FYI, you could use first here:

True, but readability takes a hit.

> Will we move it to the future marconi-json-rpc library?

Yep! I'll make another PR for that.

PR #169, opened 2023-09-27
PLT-600 Marconi providing non-existent directory in CLI option -d should create corresponding directory

brendanrbrown:
This PR:

* Adds a call to `createDirectoryIfMissing`  in `marconi-sidechain`, which is called on the parsed CLI option `-d`.
* Within `marconi-chain-index`, adds a log at the info level of the parsed CLI options for exectuables `marconi-chain-index` and `marconi-chain-index-experimental`, since
   * `marconi-sidechain` already logged the parsed CLI options, and
   *  the log will provide some possibly helpful context to the user if creating the directory fails.

Note I moved the CLI parsing in `Marconi.ChainIndex.Run.run` to be within the `Trace` context, to make it conform somewhat more closely to `Marconi.Sidechain.Run.run`.   

Pre-submit checklist:
- Branch
    - [X] Commit sequence broadly makes sense and have useful messages
    - [X] Relevant tickets are mentioned in commit messages
- PR
    - [X] Targeting main unless this is a cherry-pick backport
    - [X] Self-reviewed the diff
    - [X] Useful pull request description
    - [X] Reviewer requested


marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
```diff
@@ -20,10 +21,20 @@ import Marconi.ChainIndex.Utils qualified as Utils
...
 
 run :: Text -> IO ()
 run appName = do
+  trace <- defaultStdOutLogger appName
+
+  logInfo trace $ appName <> "-" <> Text.pack Cli.getVersion
+
   o <- Cli.parseOptions
+
+  logInfo trace . Text.toStrict $ pShowDarkBg o
```
koslambrou, line 34: What is that?

koslambrou (approved):

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
brendanrbrown: This line logs the parsed options at the info level, in a way that is consistent with `Marconi.Sidechain.Run.run`.

I assumed that in `marconi-sidechain` the use of [pShowDarkBg](https://hackage.haskell.org/package/pretty-simple-4.1.2.0/docs/Text-Pretty-Simple.html#v:pShowDarkBg) was intentional for style purposes. `pShowDarkBg` aliases `pShow`, so we could swap `pShow` in without changing the log output.

Does that answer your question, or were you referring to some other part of the line?

koslambrou: Why not just `logInfo trace $ Text.pack $ show o`? I think that's what we do in other parts.

PR #170, opened 2023-09-27
PLT-6359 Create SQL index tables only once fully synced

koslambrou:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers.hs
@@ -7,8 +7,10 @@ module Marconi.ChainIndex.Experimental.Indexers where
 
 import Cardano.Api qualified as C
 import Cardano.BM.Tracing qualified as BM
+import Control.Lens ((?~))
 import Control.Monad.Cont (MonadIO)
 import Control.Monad.Except (ExceptT, MonadError, MonadTrans (lift))
+import Data.Function ((&))
 import Data.List.NonEmpty qualified as NonEmpty
 import Data.Text (Text)
 import Data.Text qualified as Text
@@ -58,15 +60,14 @@ buildIndexers
       , -- Indexing part
         Core.WithTrace IO Core.Coordinator (WithDistance BlockEvent)
       )
-buildIndexers securityParam catchupConfig utxoConfig mintEventConfig epochStateConfig logger path = do
-  let mainLogger = BM.contramap (fmap (fmap $ Text.pack . show)) logger
-      txBodyCoordinatorLogger = BM.appendName "txBody" mainLogger
+buildIndexers securityParam catchupConfig utxoConfig mintEventConfig epochStateConfig textLogger path = do
+  let txBodyCoordinatorLogger = BM.appendName "txBody" textLogger
 
   StandardWorker blockInfoMVar blockInfoWorker <-
-    blockInfoBuilder securityParam catchupConfig mainLogger path
+    blockInfoBuilder securityParam catchupConfig textLogger path
 
   Core.WorkerIndexer _epochStateMVar epochStateWorker <-
-    epochStateBuilder securityParam catchupConfig epochStateConfig mainLogger path
+    epochStateBuilder securityParam catchupConfig epochStateConfig textLogger path
 
   StandardWorker utxoMVar utxoWorker <-
     utxoBuilder securityParam catchupConfig utxoConfig txBodyCoordinatorLogger path
@@ -94,10 +95,11 @@ buildIndexers securityParam catchupConfig utxoConfig mintEventConfig epochStateC
       UtxoQuery.mkUtxoSQLiteQuery $
         UtxoQuery.UtxoQueryAggregate utxoMVar spentMVar datumMVar blockInfoMVar
 
+  let indexerEventLogger = BM.contramap (fmap (fmap $ Text.pack . show)) textLogger
   mainCoordinator <-
     lift $
       standardCoordinator
-        mainLogger
+        indexerEventLogger
         [blockInfoWorker, epochStateWorker, coordinatorTxBodyWorkers]
 
   resumePoint <- Core.lastStablePoint mainCoordinator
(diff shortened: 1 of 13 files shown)


berewt:
I like the general idea but I'm not sure I'm comfortable using `Tracer` to add the index if we have another mean. If possible, moving to an action (or a list of actions) to trigger once we're synced would be my preference, but I'm open to discussing both solutions.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers.hs
```diff
@@ -69,13 +71,13 @@ buildIndexers securityParam catchupConfig utxoConfig mintEventConfig epochStateC
     epochStateBuilder securityParam catchupConfig epochStateConfig mainLogger path
 
   StandardWorker utxoMVar utxoWorker <-
-    utxoBuilder securityParam catchupConfig utxoConfig txBodyCoordinatorLogger path
+    utxoBuilder securityParam catchupConfig utxoConfig txBodyCoordinatorLogger textLogger path
```
koslambrou, line 74: Don't like that part of passing two different loggers here. Suggestions accepted.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Spent.hs
```diff
@@ -133,6 +133,50 @@ mkSpentIndexer path = do
...
+                  WHERE type='index' AND name='|]
+                  <> fromString indexName
+                  <> "'"
+              )
+      unless doesIndexExist $ do
+        logInfo stdoutTrace $
+          "Creating SQL index table '" <> Text.pack indexName <> "' for Spent indexer..."
+        liftIO $ SQL.execute_ c createIndexStatement
+        logInfo stdoutTrace $
+          "Finished creating SQL index table '" <> Text.pack indexName <> "' for Spent indexer..."
```
koslambrou, line 178: This is slightly duplicated across some indexers, but I don't see the ideal place to put it.

berewt: Add a helper function to either `Core.Marconi.Indexer.SQLite` or `Core.Marconi.Transformer.WithCatchup`?

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/MintTokenEvent.hs
```diff
@@ -354,15 +354,51 @@ mkMintTokenIndexer dbPath = do
...
+                  <> fromString indexName
+                  <> "'"
+              )
+      unless doesIndexExist $ do
+        logInfo stdoutTrace $
+          "Creating SQL index table '" <> Text.pack indexName <> "' for MintTokenEvent indexer..."
+        liftIO $ SQL.execute_ c createIndexStatement
+        logInfo stdoutTrace $
+          "Finished creating SQL index table '" <> Text.pack indexName <> "' for MintTokenEvent indexer..."
+
```
berewt, line 401: As `CREATE INDEX IF NOT EXISTS` already takes care of not doing anything if we have the index in place, do we want to check it manually to provide a feedback?

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Utxo.hs
```diff
@@ -158,6 +159,42 @@ mkUtxoIndexer path = do
...
+            <> " ON utxo (address)"
+    createIndexTable c addressIndexName createAddressIndexStatement
+
+    let slotNoIndexName = "utxo_slotNo"
+        createSlotNoIndexStatement =
+          "CREATE INDEX IF NOT EXISTS "
+            <> fromString slotNoIndexName
+            <> " ON utxo (slotNo)"
+    createIndexTable c slotNoIndexName createSlotNoIndexStatement
+  where
```
berewt, line 178: Minor: I really think we should avoid mixing `let` and `where`.

berewt (approved):
It looks good to me, should we handle the primary kes this way too? In this ticket or in another one.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/MintTokenEvent.hs
koslambrou: Exactly, it's just for the logging.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Spent.hs
koslambrou: I guess I can put it in `Core.Marconi.Indexer.SQLite`, but feels a bit out-of-place.

koslambrou:
> It looks good to me, should we handle the primary kes this way too? In this ticket or in another one.

The issue with the primary key is that it also checks for uniqueness, which (I guess) you want to fail as soon as possible if the constraint is violated?

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Spent.hs
berewt: In a submodule of one of these two?

berewt:
> > It looks good to me, should we handle the primary kes this way too? In this ticket or in another one.
> 
> The issue with the primary key is that it also checks for uniqueness, which (I guess) you want to fail as soon as possible if the constraint is violated?

Yes, at the same time if we're able to sync with the primary key available during a test on mainnet, there's no reason to hit a failure on a next run. We could handle it with a flag maybe?

Although, we can keep it as is in this PR and think about how to deal with this specific case in another ticket.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Indexers/Spent.hs
koslambrou: I put it in `Marconi.Core.Transformer.WithCatchup.SQLite`.

koslambrou:
> Yes, at the same time if we're able to sync with the primary key available during a test on mainnet, there's no reason to hit a failure on a next run. We could handle it with a flag maybe?

Good point! I'll create a ticket for that.

PR #184, opened 2023-10-05
PLT-7045 - Marconi sidechain exit codes during SIGINT or SIGTERM

willjgould:
When the program is closed with SIGINT or SIGTERM, we will now alter the exit code based on whether or not any indexers experience a timeout during shutdown.

I've added generalised machinery to both make these changes easier to read and to allow us to possibly re-use it, if necessary. If we do want to re-use it, we'd most likely want to move it (specifically `Concurrency.hs`) to a different file.

Additionally, this ONLY works _after_ a SIGINT or SIGTERM has been received. It feels like stepping outside the realms of that is going out of scope, because I think there's potentially a bunch more machinery that'll be needed for general exception handling (in the context of remapping exit codes).

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
@@ -43,7 +43,6 @@ import Control.Concurrent (
   newEmptyMVar,
   newMVar,
   readMVar,
-  threadDelay,
   tryPutMVar,
   tryReadMVar,
  )
@@ -62,8 +61,6 @@ import Control.Concurrent.STM.TChan (
   writeTChan,
  )
 import Control.Exception (
-  Exception,
-  SomeException (SomeException),
   catch,
   finally,
   onException,
@@ -89,7 +86,6 @@ import Control.Monad.Trans.Except (
   ExceptT,
   runExceptT,
  )
-import Data.Data (Typeable)
 import Data.Functor (($>))
 import Data.List.NonEmpty (NonEmpty)
 import Data.Map (Map)
@@ -100,9 +96,8 @@ import Data.Sequence qualified as Seq
 import Data.Text qualified as Text
 import Data.Void (Void)
 import Data.Word (Word64)
-import GHC.Exception.Type (Exception (fromException), toException)
 import Marconi.ChainIndex.Error (
-  IndexerError (CantInsertEvent, CantRollback, CantStartIndexer),
+  IndexerError (CantInsertEvent, CantRollback, CantStartIndexer, Timeout),
   ignoreQueryError,
  )
 import Marconi.ChainIndex.Indexers.AddressDatum (
@@ -144,7 +139,6 @@ import Prometheus qualified as P
 import Streaming qualified as S
 import Streaming.Prelude qualified as S
 import System.Directory (createDirectoryIfMissing)
-import System.Exit (ExitCode (ExitFailure), exitWith)
 import System.FilePath (
   takeDirectory,
   (</>),
@@ -739,11 +733,7 @@ runIndexers
             waitQSemN (coordinator ^. barrier) (coordinator ^. indexerCount)
         case res of
           Just _ -> logInfo stdoutTrace "Done!"
-          Nothing -> throwIO TimeoutException
-
-data IndexerException = TimeoutException
-  deriving (Show, Typeable)
-instance Exception IndexerException
+          Nothing -> throwIO (Timeout @Void "Timed out.")
 
 updateProcessedBlocksMetric
   :: S.Stream (S.Of (ChainSyncEvent BlockEvent)) IO r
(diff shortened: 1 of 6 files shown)


marconi-sidechain/src/Marconi/Sidechain/Concurrency.hs
```diff
@@ -0,0 +1,107 @@
...
+  If the program is exiting due to a SIGINT or SIGTERM and if we've got an MVar in our result,
+  we set the exit code to whatever it specifies. If either of the previous points aren't true,
+  we exit with whatever signal (SIGINT or SIGTERM) requested shutdown. -}
+
+{- | A data type representing an action with explicit instructions as to whether concurrent
+		 operations should attempt to map any exceptions it throws to Unix exit codes.
+
+		It has two type parameters:
+
+		* 'e' the exception type, 'Void' in the unhandled case
```
berewt, line 41: I think Haddock don't like single quotes for type parameters, as it tries to resolve them to a known function. You probably want to use `@a@` and `@e@`.

marconi-sidechain/src/Marconi/Sidechain/Concurrency.hs
```diff
@@ -0,0 +1,107 @@
...
+  Handled :: (Exception e, HasExitCode e) => IO a -> HandledAction e a
+  Unhandled :: IO a -> HandledAction Void a
+
+{- | Perform two actions, with explicit handling instructions, concurrently (with the chosen scheme)
+
+		Takes:
+
+		* 'concOp' - A concurrent operation (`race` or `concurrent`)
+		* 'left' - The "left" `HandledAction` for the concurrent operation to run
+		* 'right - The "right" `HandledAction` for the concurrent operation to run
```
berewt, line 55: Move this comment directly next to the parameter using `-- |` (above) or `-- ^` (below) the parameter declaration.
We lose the name info but it isn't included in the signature anyway.

marconi-sidechain/src/Marconi/Sidechain/Error.hs
```diff
@@ -16,3 +29,48 @@ data QueryExceptions
...
   deriving stock (Show)
   deriving anyclass (Exception)
+
+-- * Mapping of exceptions to exit codes
+
+class HasExitCode e where
+  toExitCode :: e -> Int
+
+instance HasExitCode (IndexerError Void) where
+  toExitCode = indexerErrorToExit
```
berewt, line 39: Same here, `Void` can be replaced by `a`.

marconi-sidechain/src/Marconi/Sidechain/Error.hs
```diff
@@ -16,3 +29,48 @@ data QueryExceptions
...
+		Maps SIGINT (2) to 130
+		Maps SIGTERM (15) to 143
+		Maps all other signals to 137 (SIGKILL)
+-}
+signalToExit :: Int -> Int
+signalToExit signal =
+  case signal of
+    2 -> 130
+    15 -> 143
+    _ -> 137
```
berewt, line 76: Why didn't you take a `Signal` as an input, together with guards, it could lead to the following

```
signalToExit :: Signal -> Int
signatToExit signal
  | signal == sigINT = 130
  | signal == sigTERM = 143
  | otherwise = 137
```

(we might event introduce constant to provide semantic for `130`, `143` and `137`)

marconi-sidechain/src/Marconi/Sidechain/Run.hs
```diff
@@ -30,6 +31,9 @@ import Text.Pretty.Simple (pShowDarkBg)
 * marconi indexer workers
 
 Exceptions in either thread will end the program
+
+If the program is terminated with SIGINT or SIGTERM, exceptions in the marconi indexer workers
+thread will be mapped to exit codes
```
berewt, line 36: It would be useful to add a link to the `signalToInt` function or to describe the mapping in the comment.

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -723,10 +728,12 @@ runIndexers
...
             <> Text.pack (show secondsBeforeTimeout)
             <> "s) ..."
-        void $
+        res <-
           timeout (secondsBeforeTimeout * 1_000_000) $
             waitQSemN (coordinator ^. barrier) (coordinator ^. indexerCount)
-        logInfo stdoutTrace "Done!"
+        case res of
+          Just _ -> logInfo stdoutTrace "Done!"
+          Nothing -> throwIO (Timeout @Void "Timed out.") -- TODO: What useful info can go here?
```
berewt, line 736: It's hard to tell, as we don't know which indexer did timeout here.
It's likely that in the new indexer implementation, we'll try to move the timeout in each worker, to attach the worker name to the error message.

berewt (approved):
It looks great. A follow-up would be to check if the timeout is handled properly in the new implementation and to replicate the behaviour.
I'll create the ticket for it.

marconi-sidechain/src/Marconi/Sidechain/Error.hs
willjgould: Did you mean to add another comment at a different point where I've used `Void`? I'm assuming you mean line 41 of this file, but I just want to double check.

PR #186, opened 2023-10-06
PLT-7794 - Check if file exists and strictly evaluate error textification

willjgould:
There was a race condition between the mapping of errors returned by `buildIndexers` to `Text` and the exiting of the program with a failure condition. 

~I've resolved it by strictly evaluating the mapping.~

I've resolved it by using the `shutdown` function from `Cardano.BM`. This tells the logger to shutdown, and also waits for the shutdown to complete. The shutdown involves draining the queue, guaranteeing our message is processed.

Also, I'm now checking if the specified JSON file exists at the top level before continuing, which isn't strictly necessary as a part of this ticket now that I've fixed the root cause, but it makes sense so I'm leaving it in.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 
@@ -6,6 +7,7 @@ module Marconi.ChainIndex.Experimental.Run where
 import Cardano.Api qualified as C
 import Cardano.BM.Trace (logError, logInfo)
 
+import Control.Monad (unless)
 import Control.Monad.Except (runExceptT)
 import Data.Text (Text)
 import Data.Text qualified as Text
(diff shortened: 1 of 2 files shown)


marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
```diff
@@ -67,7 +69,10 @@ run appName = withGracefulTermination_ $ do
       preferredStartingPoint = Cli.optionsChainPoint $ Cli.commonOptions o
 
   nodeConfigPath <- case Cli.optionsNodeConfigPath o of
-    Just cfg -> pure cfg
+    Just cfg -> do
+      exists <- doesFileExist cfg
+      unless exists $ error ("Config file does not exist at the provided path: " <> cfg)
```
koslambrou, line 74: Use `logError` instead for printing the error message (for consistent logging), followed by `exitFailure`.

koslambrou (approved):

willjgould:
I've reworked the changes with what I think is a much more correct solution. The problem occurred because we weren't waiting for the logger queue to finish processing. This is resolved by using the `shutdown` function, which terminates the dispatcher after waiting for the queue to drain, guaranteeing our errors will be processed before exit.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Run.hs
```diff
@@ -66,9 +68,32 @@ run appName = withGracefulTermination_ $ do
...
+      asynchronously.
+
+      In order to ensure that we always get errors written to the console, we need to wait for the
+      queue to complete before exiting. We can use the 'shutdown' function to guarantee this.
+
+      'shutdown' puts a kill-pill on the switchboard's queue, then waits for the switchboard's
+      dispatcher to exit, which requires the queue to be drained, ensuring all messages are
+      processed.
+  -}
+  let informativeFailure :: Text -> IO a
```
koslambrou, line 82: informativeFailure -> logInformativeFailure? Or maybe just logFullError?

willjgould: Hmmm, it was because I'm also calling `exitFailure` here. Maybe I should just remove the `exitFailure` call and use your naming suggestion.

PR #190, opened 2023-10-10
PLT-7861 Fix resuming and lastSync for the EpochState indexer

berewt:
Many steps were missing to make it works:
- Catchup didn't mix well with sparse events when no event were issued, we always store the event in the batch and thus we weren't updating the last sync point. As a consequence, the SDD and nonce indexer were updating their last sync point only at the end of epoch, causing the `queryLatest` call to fail with a `AheadOfSync` error.
  The catchup mechanism is still the same, but the standard indexer was updated to always forward the distance even when we produce an empty event.
- We didn't use the nodeToClient encoding of blocks, and consequently storing invalid block information. Especially the block header hash was wrong. The code was changed to use `NodeToClient` encoding/decoding.

Note that we may want to revisit the catchup mechanism later, but I considered it as out of scope for this PR.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


koslambrou (approved):
LGTM.

How did you notice the issue of the wrong storage of block information? What this related to the rollback issue?

PR #180, opened 2023-10-04
PLT-7932 Add tracking of the chain tip

berewt:
- The processing of the chain-sync events is modified to issue some events with the chain tip.
- We introduce the `LastEventIndexer`, an indexer that saves the last value of an event on disk, to use it to save chain tips.
- We modify the indexers organisation accordingly.

Pre-submit checklist:
- Branch
    - [ ] Tests are provided (if possible)
    - [ ] Commit sequence broadly makes sense and have useful messages
    - [ ] Important changes are reflected in changelog.d of the affected packages
    - [ ] Relevant tickets are mentioned in commit messages
- PR
    - [ ] (For external contributions) Corresponding issue exists and is linked in the description
    - [ ] Targeting main unless this is a cherry-pick backport
    - [ ] Self-reviewed the diff
    - [ ] Useful pull request description
    - [ ] If relevant, reference the ADR in the PR and reference the PR in the ADR
    - [ ] Reviewer requested

PR #200, opened 2023-10-16
PLT-8028: remove initial rollback

ana-pantilie:
From my discussion with @berewt:

https://github.com/input-output-hk/marconi/pull/180 introduces the following [indexing event](https://github.com/input-output-hk/marconi/blob/3f5750d9ac0451618fa4893731a5be3196fed0ed/marconi-chain-index/src/Marconi/ChainIndex/Experimental/Runner.hs#L207) when processing a rollback. This ends up triggering https://github.com/input-output-hk/marconi/blob/3f5750d9ac0451618fa4893731a5be3196fed0ed/marconi-core/src/Marconi/Core/Preprocessor/Resume.hs#L86. The solution is to remove the "artificial" rollback event that `marconi` emits, which is no longer necessary for synchronizing indexers.

Note: the `ResumeStarts` state, which was introduced to handle rollbacks at the start of the resuming procedure, could not be removed because there is an additional initial rollback which comes from somewhere external to `marconi`. 

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


berewt (approved):
Looks good to me.

@koslambrou it looks like we still have an initial rollback after we remove the forged one created when the intersection is found. Should we open a spike to investigate the why and decide whether we remove or keep it?

koslambrou:
> Should we open a spike to investigate the why and decide whether we remove or keep it?

Looking at the code... I think that If we get 2 rollbacks, that means there is an initial one we emit when the intersection is found, and the second is an actual rollback from the node that tells the client to go back some previous state.

PR #201, opened 2023-10-17
PLT-8058 - Added getBurnTokenEvents endpoint, defunctionalised MintTokenEventsMatchingQuery

willjgould:
I've added a `getBurnTokenEvents` endpoint that returns its data with a `stability` tag.

I've also copied the `GetBurnTokenEvents` type and machinery from the sidechain endpoints.

I've also defunctionalised `MintTokenEventsMatchingQuery` so that we can expose it, if we choose to.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


marconi-chain-index/src/Marconi/ChainIndex/Experimental/Api/HttpServer.hs
```diff
@@ -127,6 +148,81 @@ getEpochNonceHandler
...
+            stable
+            <$> NonEmpty.toList (timed ^. Core.event . mintTokenEvents)
+
+    {- Map an internal event to a single component of our 'getBurnTokenEventsHandler' response
+      object -}
+    mapEvent :: C.ChainPoint -> Bool -> MintTokenEvent.MintTokenEvent -> BurnTokenEventResult
+    mapEvent chainPoint isStable event =
+      let (slotNo, headerHash) = case chainPoint of
+            (C.ChainPoint slot hash) -> (slot, hash)
+            C.ChainPointAtGenesis -> (0, "genesis")
```
willjgould, line 214: I'm not sure about this bit in particular, is it correct?

berewt: There shouldn't be any minting/burning at genesis. Although, for completeness, I would probably prefer to use `Maybe SlotNo` and `Maybe BlockHeaderHash` rather than forging values.

berewt (commented):
I'd suggest to move the content added to `Marconi.Experimental.Api.HttpServer` and `Marconi.Experimental.Api.Routes` to its own sub module, probably `Marconi.Expermental.Api.Endpoint.MintBurnToken`?
We tried in the past to put all the stuff in routes and server and in the end it's a huge mess that is hard to maintain.

berewt (approved):
Very great job overall, it's clean and concise.
I'm not sure about the JSON-RPC/Rest split. Although, it's a concern for a few endpoints so it should be ok.
To be sure, I'd suggest double-checking with either @koslambrou or @neilmayhew who both worked on the HTTPServer recently to check if the module structure is ok for them.

neilmayhew:
> Very great job overall, it's clean and concise.

I second that!

> I'm not sure about the JSON-RPC/Rest split. Although, it's a concern for a few endpoints so it should be ok. To be sure, I'd suggest double-checking with either @koslambrou or @neilmayhew who both worked on the HTTPServer recently to check if the module structure is ok for them.

I like the split. The REST and RPC endpoints are quite different from one another, so it's helpful to have them separated. Also, I don't think we're likely change any of the REST endpoints or add new ones, so it's less distracting to have them off to the side somewhere.

I also like breaking each endpoint into its own module. Although that seems over the top for the smaller ones (eg `echo`!) some of the others are getting quite big and it's helpful to treat them all consistently, ie one endpoint is one module, because that makes things easier to find.

marconi-chain-index/test/Spec/Marconi/ChainIndex/Experimental/Api/Routes.hs
```diff
@@ -0,0 +1,68 @@
...
+      <$> (Just <$> Gen.genSlotNo)
+      <*> (Just <$> Gen.genHashBlockHeader)
+      <*> Gen.genBlockNo
+      <*> CGen.genTxId
+      <*> pure (fmap C.hashScriptDataBytes hsd)
+      <*> pure (fmap C.getScriptData hsd)
+      <*> CGen.genAssetName
+      <*> Gen.genQuantity (Range.linear 0 10)
+      <*> pure True
+  tripping r Aeson.encode Aeson.eitherDecode
```
neilmayhew, line 68: Would it be worth having (possibly orphan) `Arbitrary` instances for these types instead of putting the generation inside the property? We might end up using them in other tests.

willjgould: I'm very new to property testing so I'm not 100% certain, but I don't think `Hedgehog` uses `Arbitrary`, so I've just abstracted the generators and moved them into their own file.

PR #205, opened 2023-10-19
PLT-7497: introduce MarconiTrace

ana-pantilie:
Main changes:
- introduces `MarconiTrace` inside `marconi-chain-index` with which we avoid having to render to text every time we log
- changes are propagated to `marconi-chain-index-experimental`, `marconi-sidechain`, `marconi-tutorial` and their respective testing packages
- reverses the dependency of the `Types` module on the `Retry` module (`Types` should be higher up the hierarchy), effectively this just moves the definition of the config retry type to `Types`

I have tested this manually and logging seems to work fine, as before.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


neilmayhew (changes requested):
I like what you've done here. Well done for figuring out that you can use a contramapping approach.

At first I was concerned that we wouldn't have a way to do logging with plain `Text` but I'm not sure now why we would want to do that. Literal text will be converted to `Doc` automatically via the `IsString` instance, and if we want to output a non-literal block of `Text` we can just use `pretty` on it.

I also think it's good to make the use of `MonadReader` a separate PR (one that we may well want to leave for later).

As for using `prettyprinter-interp`, I've re-read the documentation and I now see that it's a thin wrapper around `string-interpolate` which uses `Show` instances when you interpolate, and we need one that uses `Pretty` instances. I wasn't immediately able to find a quasi-quoter that uses `Pretty` so this is probably another thing we want to leave for later.

neilmayhew:
FYI, it's OK to rebase a PR branch to bring in changes from `main` rather than merging `main` into it, especially if no-one's reviewed it yet.

We also do squashing and rebasing before merging finished PRs, and the repo options require fast-forward merges. Typically we squash into a single commit, but it's also OK to have two or three commits if that makes better sense. We do, however, like to have the ticket ID in the commit message of every merged commit, so we can see which commits were part of a given ticket afterwards. Since we use fast-forward merges, there's no merge commit to show which PR commits came from.

ana-pantilie:
Thank you @neilmayhew ! Great catch with the extra type param, I really didn't like that I had to introduce that, but I wasn't aware it's not necessary now.

Related to your last comment, do you want me to clean-up the history of this PR? I was planning to just squash it to one commit. Sorry if the commit history was messy when you reviewed it, I will try to improve that in the future!

neilmayhew:
> … do you want me to clean-up the history of this PR? I was planning to just squash it to one commit. Sorry if the commit history was messy when you reviewed it, I will try to improve that in the future!

No worries! The commit history is perfectly fine as it is, and I didn't find it confusing. Squashing at the end is also fine and expected.

My main intention was to let you know that rebasing is considered acceptable by the the longer-serving team members (although I'm sure they're fine with merges from `main`, too).

Also, I wasn't sure if anyone had explained the merging policies (since I found out only by accident) so I thought it might be helpful to mention them.

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Runner.hs
```diff
@@ -73,8 +73,8 @@ data RunIndexerEventPreprocessing event = RunIndexerEventPreprocessing
 Lens.makeLenses ''RunIndexerEventPreprocessing
 
 -- | Common configuration required to run indexers
-data RunIndexerConfig event = RunIndexerConfig
-  { _runIndexerConfigTrace :: Trace.Trace IO Text
+data RunIndexerConfig event ann = RunIndexerConfig
```
neilmayhew, line 76:
```suggestion
data RunIndexerConfig event = RunIndexerConfig
```

marconi-chain-index/src/Marconi/ChainIndex/Experimental/Runner.hs
```diff
@@ -108,7 +108,7 @@ runIndexer
      , Core.Closeable IO indexer
      , Core.Point a ~ C.ChainPoint
      )
-  => RunIndexerConfig a
+  => RunIndexerConfig a ann
```
neilmayhew, line 111:
```suggestion
  => RunIndexerConfig a
```

neilmayhew (approved):
This is looking good. There are just a couple of places where the `ann` type parameter still needs to be removed, and then it's good to go. I'll approve this now so you don't need to get another review from me when you've made those changes, and you can go ahead and merge when you're ready.

neilmayhew:
> I'll approve this now so you don't need to get another review from me when you've made those changes, and you can go ahead and merge when you're ready.

However, I'm happy to look at it again if you'd like me to. I see that there are merge conflicts, so if you have any concerns just let me know.

ana-pantilie:
@neilmayhew Ah, can't believe I missed those, thank you for catching them!

PR #212, opened 2023-10-24
PLT-7497: restructure Marconi with a focus on logging

ana-pantilie:
Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested

### DIFF
--- a/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
+++ b/marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
@@ -86,6 +86,7 @@ import Control.Monad.Trans.Except (
   ExceptT,
   runExceptT,
  )
+import Control.Monad.Trans.Reader (ReaderT (runReaderT))
 import Data.Functor (($>))
 import Data.List.NonEmpty (NonEmpty)
 import Data.Map (Map)
@@ -110,9 +111,10 @@ import Marconi.ChainIndex.Indexers.EpochState qualified as EpochState
 import Marconi.ChainIndex.Indexers.MintBurn qualified as MintBurn
 import Marconi.ChainIndex.Indexers.ScriptTx qualified as ScriptTx
 import Marconi.ChainIndex.Indexers.Utxo qualified as Utxo
-import Marconi.ChainIndex.Logging (chainSyncEventStreamLogging)
+import Marconi.ChainIndex.Logging (chainSyncEventStreamLogging, logMInfo)
 import Marconi.ChainIndex.Node.Client.Retry (withNodeConnectRetry)
 import Marconi.ChainIndex.Types (
+  ChainIndexerT (ChainIndexerT),
   IndexingDepth (MaxIndexingDepth, MinIndexingDepth),
   RunIndexerConfig (RunIndexerConfig),
   SecurityParam (SecurityParam),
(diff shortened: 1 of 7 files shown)


marconi-chain-index/src/Marconi/ChainIndex/Logging.hs
```diff
@@ -54,6 +59,14 @@ marconiFormatting =
   Pretty.renderStrict
     . Pretty.layoutPretty Pretty.defaultLayoutOptions
 
+logMInfo :: MonadIO m => Pretty.Doc () -> ChainIndexerT m ()
+logMInfo msg = do
+  trace <- Lens.view runIndexerConfigTrace
+  lift $ logInfo trace msg
```
neilmayhew, line 65: You could make this into a helper function that you use to map the other log functions:

```suggestion
logToLogM :: MonadIO m => (MarconiTrace m -> Pretty.Doc () -> m ()) -> Pretty.Doc () -> ChainIndexerT m ()
logMInfo logf msg = do
  trace <- Lens.view runIndexerConfigTrace
  lift $ logf trace msg

logMInfo :: MonadIO m => Pretty.Doc () -> ChainIndexerT m ()
logMInfo = logToLogM logInfo
```

marconi-chain-index/src/Marconi/ChainIndex/Logging.hs
```diff
@@ -54,6 +59,14 @@ marconiFormatting =
...
+
+-- TODO: this should also throw an exception
+logMError = undefined
```
neilmayhew, line 68: I feel like log functions shouldn't be throwing exceptions, despite what we said earlier today about having a unified framework for both.

Perhaps it would be OK to have a function that does both, but have that reflected in the name. Something like `logThrowError`.

In another system I worked on, we also included a "context" parameter which was just a string that indicated where the error was thrown from. This helped a lot when looking at logs.

marconi-chain-index/src/Marconi/ChainIndex/Node/Client/Retry.hs
```diff
@@ -33,6 +33,9 @@ data RetryState = RetryState
   , secondsBeforeNextRetry :: !Word64
   }
 
+-- TODO: this function is problematic because it is used both inside the
+-- ChainIndexT context and outside to initialize the config
+-- see Marconi.ChainIndex.Run.run
```
neilmayhew, line 38: Could you make the monad be a type parameter `m` and give it constraints `(MonadIO m, MonadUnliftIO m)`?

Or have two versions of the function, one of which wraps the other and lives in `ChainIndexT`?

marconi-chain-index/src/Marconi/ChainIndex/Types.hs
```diff
@@ -65,6 +73,17 @@ import Database.SQLite.Simple.ToField qualified as SQL
 import GHC.Generics (Generic)
 import Prettyprinter (Doc)
 
+newtype ChainIndexerT m a = ChainIndexerT
+  {getChainIndexerT :: ReaderT (RunIndexerConfig m) m a}
+  deriving newtype (Functor, Applicative, Monad, MonadReader (RunIndexerConfig m))
+
+instance MonadTrans ChainIndexerT where
+  lift = ChainIndexerT . lift
```
neilmayhew, line 81: Maybe you need a `MonadIO` instance too

marconi-chain-index/src/Marconi/ChainIndex/Types.hs
```diff
@@ -142,13 +161,15 @@ newtype TxIndexInBlock = TxIndexInBlock Word64
     )
 
 -- | Common configuration required to run indexers
-data RunIndexerConfig = RunIndexerConfig
-  { _runIndexerConfigTrace :: MarconiTrace IO
+data RunIndexerConfig m = RunIndexerConfig
+  { _runIndexerConfigTrace :: MarconiTrace m
```
neilmayhew, line 165: Maybe you could leave this with `IO` and use `liftIO` to run it? It seems a little odd to have the config parameterized. I feel like it should be possible to use the same config with any `MonadIO`. The version of the logging functions that operate in the `ReaderT` could do the `liftIO` so that the caller doesn't have to.

neilmayhew (commented):
My comments and suggestions here are fairly random and untested, so feel free to ignore them if they don't seem like they will work

marconi-chain-index/src/Marconi/ChainIndex/Logging.hs
ana-pantilie: Great idea, thanks!

marconi-chain-index/src/Marconi/ChainIndex/Logging.hs
ana-pantilie: I agree, I think we need to have the concrete requirements regarding errors and exceptions in order to make any changes here.

marconi-chain-index/src/Marconi/ChainIndex/Types.hs
ana-pantilie: I'll try that, thanks for pointing it out!

marconi-chain-index/src/Marconi/ChainIndex/Node/Client/Retry.hs
ana-pantilie: I'm not sure about using `MonadUnliftIO`. If I understand correctly, it looks like we'd be restricting our `ChainIndexT` monad stack to just having reader capabilites, and I don't have enough knowledge about the domain if we need to worry about this or not.

ana-pantilie: Hmm, still, for the second solution I believe you end up explicitly implementing the `ChainIndexT -> IO` transformation. Unfortunately this isn't the only place in the code this transformation is needed _I think_, not 100% sure yet but wanted to add to my previous comment.

neilmayhew (commented):
This is coming together nicely

marconi-chain-index/src/Marconi/ChainIndex/Indexers.hs
```diff
@@ -638,94 +651,104 @@ mkIndexerStream' f coordinator =
...
+                <> "Are you sure you want to restart syncing that indexer from genesis?"
+                <> line
+                <> "If so, remove the '--fail-if-resyncing-from-genesis' flag."
+            )
+      else do
+        let stream =
+              mkIndexerStream coordinator
+                . chainSyncEventStreamLogging
+                . updateProcessedBlocksMetric
+            runChainSyncStream = liftIO $ withChainSyncBlockEventStream socketPath networkId [resumePoint] stream
```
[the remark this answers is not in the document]
ana-pantilie: ## Overview of one of the problems encountered in my investigation

`chainSyncEventStreamLogging` needs to have access to the logging context.

The `stream` is problematic here because it needs to both be in the `ChainIndexT` context (as noted above) and in `IO` because it is run by `withChainSyncBlockEventStream`. Inside, it uses `withAsync` on the stream consumer which means that we would need a way to "unlift" the `ChainIndexT` back to `IO`.

This brings forward the question of whether we can make `ChainIndexT` an instance of `MonadUnliftIO`. This would restrict the monad stack to allow only the current reader capability, and also it does not play well with the `ExceptT` we have decided to keep (as per @willjgould's investigation).

I believe that there are two possible ways to go on from here:
1. we make `ChainIndexT` an instance of `MonadUnliftIO` and see how it impacts the usage of `ExceptT`, keeping in mind the restrictions we will impose
2. we refactor the code in such a way that rids it of how deeply coupled it is with IO now, as Will also suggested; I don't know if this is possible, someone with more knowledge of the business logic should have a better idea than me

PR #214, opened 2023-10-25
Add a component diagram to the RTD architecture documentation

koslambrou:
There are two diagrams in the page.

First the repository structure:

image

Second the decision tree:

image

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


berewt (approved):
I'd add a story to provide text information that replicates the content of the diagram, to both provide extra details if needed and for accessibility purposes.

koslambrou:
@berewt I can do that as part of this story. However, isn't the diagram pretty self-explanatory? Each box and link has some text describing what the block and the relation is about.

berewt:
To my eyes it is but:

1.  I don't know whether it's clear enough for external contributors (although, we can add it later).
2. Even if the intersection of Marconi users and visually impaired people is likely to be empty, it's still good to provide a text alternative.

(again, it's probably not urgent)

ana-pantilie:
@koslambrou which executables are these? Are the diagrams based on a renaming we're going to do?

koslambrou:
> @koslambrou which executables are these? 

@ana-pantilie There is only one executable: it's `marconi-chain-index`. 

> Are the diagrams based on a renaming we're going to do?

Yes, these diagrams are based on the renaming we are planning to do.

>  Also, from the diagrams it seems to me like if you want RPC capabilities then you end up losing the Cardano specific functionality?

Can you elaborate on why that is the case?

koslambrou:
@ana-pantilie @berewt As suggested, I added much more text to explain the main diagram.

ana-pantilie:
> Can you elaborate on why that is the case?

I guess the idea is that if you want both Cardano and JSON-RPC you use both packages, correct? Maybe this should be made explicit because I got a bit confused by the diagram, I was expecting it to tell me which single solution to use not how to pick a combination of solutions.

koslambrou:
@ana-pantilie Ahh, you're referring to the 2nd diagram?

ana-pantilie:
@koslambrou yes

PR #215, opened 2023-10-25
PLT-7727 migrate legacy marconi chain index

brendanrbrown:
Note `marconi-sidechain` used the legacy chain index package, and `marconi-starter` used the experimental (current) one. That is now explicit in the imports.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ ] Reviewer requested


legacy/marconi-chain-index-legacy/cardano-api-extended-legacy/src/Cardano/Api/Extended.hs
```diff
@@ -0,0 +1,24 @@
+{-# OPTIONS_GHC -Wno-missing-import-lists #-}
```
koslambrou, line 1: Doesn't change anything I believe, but couldn't you just refer to the `cardano-api-extended` of `marconi-chain-index`

koslambrou (approved):

koslambrou:
@sjoerdvisscher I would suggest for this PR to merge and rebase your branch on `main` so that we don't lose the changes following this huge restructuring.

legacy/marconi-chain-index-legacy/cardano-api-extended-legacy/src/Cardano/Api/Extended.hs
brendanrbrown: I thought of that, but I was thinking it was possible we would want to add or modify `cardano-api-extended` in `marconi-chain-index` with breaking changes, which could require us to then update the `marconi-chain-index-legacy` code. It seemed better to avoid that development burden if it was not necessary.

But we can make a follow-up ticket to drop the `cardano-api-extended-legacy` and just refer to the original, if that's easier.

PR #216, opened 2023-10-25
PLT-7811 - Error logging machinery to WithTracer.hs; raise error on file write timeout during indexer close

willjgould:
The logging machinery we decided on as a result of https://input-output.atlassian.net/browse/PLT-7809 has been added.

Additionally, when a timeout occurs during async file writing in `FileIndexer` on shutdown, an error is raised and logged.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


brendanrbrown:
Noting that the merge conflicts come from the recent restructuring to migrate the "legacy" marconi-chain-index and keep only the formerly "experimental" version.

So in other words, the conflicts should be able to be resolved by just taking the change in a path `marconi-chain-index/src/Marconi/ChainIndex/Experimental/<path>` and applying it to `marconi-chain-index/src/Marconi/ChainIndex/<path>`.

marconi-core/src/Marconi/Core/Transformer/WithTracer.hs
```diff
@@ -463,3 +477,29 @@ instance
...
+  -> IndexerError
+  -> m ()
+logIndexerError = logError IndexerFailed [SomePrism _StopIndexer]
+
+-- | A helper that allows us to add an exclusion list of error constructors
+logError
+  :: forall m e' e
+   . (MonadIO m)
+  => (e' -> e)
+  -> [SomePrism e']
```
berewt, line 497: While the variable name makes the purpose explicit, can you also add variable documentation here?

marconi-core/src/Marconi/Core/Transformer/WithTracer.hs
```diff
@@ -463,3 +477,29 @@ instance
...
+  -> Trace m e
+  -> e'
+  -> m ()
+logError fromErr exclusionList tr err =
+  ( when (any (\(SomePrism p) -> isn't p err) exclusionList) $
```
berewt, line 502: Is it wrong? I read it as "when one error of the exclusion list isn't my error, log my error" (which will always be the case if we provide 2 different errors in the exclusion list).

```suggestion
  ( unless (any (\(SomePrism p) -> is p err) exclusionList) $
```

berewt (approved):

berewt (changes requested):
Sorry, I approved to fast: the error filter must be fixed (unless I misinterpreted it)

marconi-core/src/Marconi/Core/Transformer/WithTracer.hs
willjgould: Good catch!

berewt (approved):

PR #224, opened 2023-10-31
PLT-8339: Fix SQL instances for ChainPoint and ChainTip

ana-pantilie:
Both `ChainPoint` and `ChainTip` can refer to the genesis block, in this case they don't carry any data. This should be reflected in the SQL representation correctly, therefore we need to make the changes in this PR.

@berewt pointed out the tests in https://github.com/input-output-hk/marconi/blob/830f9ea0cf4834b110822df178dca39a7edf4e54/marconi-chain-index/test/Spec/Marconi/ChainIndex/Orphans.hs#L103, unfortunately I wasn't able to write similar ones for these changes. The reason is that these would be tests for the `RowParser`, and the SQL package we use doesn't seem to have a way to run this parser in a pure context. I am 100% open to suggestions, let me know if there's something I'm missing here!

Also, there doesn't seem to be a way to return errors from the `RowParser` context, so I had to throw exceptions. Please let me know if we need to do something else here.

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


berewt (approved):
LGTM.
It's ok to throw errors, we catch them anyway https://github.com/input-output-hk/marconi/blob/main/marconi-core/src/Marconi/Core/Indexer/SQLiteIndexer.hs#L213-L220 
We could maybe unwrap the `RowParser` using an internal module, but we can skip it at the moment.

PR #228, opened 2023-11-02
PLT-8078: Add missing data to new sync block endpoint

ana-pantilie:
This PR fixes the new sync block API meaning that now it should correspond to the one in marconi-sidechain, see the modified golden test output. 

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


marconi-chain-index/src/Marconi/ChainIndex/Api/JsonRpc/Endpoint/CurrentSyncedBlock.hs
```diff
@@ -42,8 +43,30 @@ data GetCurrentSyncedBlockResult = GetCurrentSyncedBlockResult
...
+          , "blockHeaderHash" .= bh
+          , "slotNo" .= sn
+          , "epochNo" .= en
+          ]
+            <> jsonNodeTip
+  toJSON (GetCurrentSyncedBlockResult _ _ _ tip _ _) =
+    Aeson.object $
+      case tip of
+        Nothing -> []
+        Just tip' -> ["nodeTip" .= toJSON tip']
```
berewt, line 69: While they may be good in the specific case of the translation from the original type, they don't allow roundtrip. Would you be ok to either refactor them so that it becomes the case (I guess you'll have to generalise what you did for `nodeTip` for the other fields) or throw an error. Or, as we use a custom anyway, put all the fields that must be defined together in a dedicated type.

berewt (commented):
I think it would worths it to make the roundtrip work.

marconi-chain-index/src/Marconi/ChainIndex/Api/JsonRpc/Endpoint/CurrentSyncedBlock.hs
ana-pantilie: Not sure I fully understand, can you be a bit more specific regarding the roundtrip you're talking about?

I'm not a very big fan of this representation either, but I was wary about refactoring this completely since I wasn't sure why it was defined as a tuple of Maybe's instead of a Maybe of a tuple initially.

berewt: The idea is that you want to have `fromJSON . toJSON` equals `pure`. Here, if some fields are `Nothing` and others are `Just`, this property doesn't hold.
The main reason for a tuple of maybes was to allow an automatic derivation of `FromJSON` and `ToJSON` but if we drop it, we can definitely move back to a `Maybe` of a tuple.

berewt (approved):
LGTM.

PR #240, opened 2023-11-16
Improve build and test instructions in marconi-sidechain README

neilmayhew:
* Update Nix prerequisites
* Put Nix first
* Add Docker instructions
* Mention `marconi-sidechain/examples/run-pre-prod-queries` for testing RPC
* Mention that Nix works on Mac

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -20,10 +20,12 @@ packages: legacy/marconi-core-legacy
           marconi-chain-index
           marconi-core
           marconi-core-json-rpc
-          doc/read-the-docs-site
           marconi-sidechain
           marconi-sidechain-experimental
           marconi-starter
+if(!os(darwin) || impl(ghc >= 9.4))
+  packages:
+          doc/read-the-docs-site
 
 -- We never, ever, want this.
 write-ghc-environment-files: never
--- a/marconi-sidechain/README.md
+++ b/marconi-sidechain/README.md
@@ -1,18 +1,29 @@
 # marconi-sidechain
 
-`marconi-sidechain` is a lightweight chain follower application for the Sidechain project to index and query specific information from the Cardano blockchain.
+`marconi-sidechain` is a lightweight chain-follower application for the Sidechain project to index and query specific information from the Cardano blockchain.
 The interface for querying the indexed information uses [JSON-RPC](https://www.jsonrpc.org/specification) over HTTP.
 
-See the [architecture documentation](./doc/ARCHITECTURE.adoc) for more information on how this application was build.
+See the [architecture documentation](./doc/ARCHITECTURE.adoc) for more information on how this application was built.
 
 ## Prerequisites
 
-If using `Nix`:
+If using **Nix**:
 
 * [Nix](https://nixos.org/download.html) (`>=2.5.1`)
-  * Enable IOHK's binary cache or else you will build the world! Refer to [this section](../CONTRIBUTING.adoc#how-to-get-a-shell-environment-with-tools) on how to achieve this.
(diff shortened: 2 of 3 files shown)


marconi-sidechain/README.md
```diff
@@ -1,37 +1,67 @@
...
+### Nix build
 
-TBD
+The `marconi-sidechain` executable is available as a nix flake.
+
+If inside the `marconi` repository, you can run from the top-level:
+
+```
+$ nix build .#marconi-sidechain
+```
```
berewt, line 42: On Mac apple silicon, we need to force a x86 build using `--system x86_64-darwin`. I didn't find a nice way to put it but if could be a nice addition.

marconi-sidechain/README.md
```diff
@@ -1,37 +1,67 @@
...
-If using `Nix`:
+If using **Nix**:
 
 * [Nix](https://nixos.org/download.html) (`>=2.5.1`)
-  * Enable IOHK's binary cache or else you will build the world! Refer to [this section](../CONTRIBUTING.adoc#how-to-get-a-shell-environment-with-tools) on how to achieve this.
+* Ensure you have the following configuration (eg in `~/.config/nix/nix.conf`):
 
-If *not* using `Nix`:
+```
+experimental-features = nix-command flakes
```
berewt, line 16: do we want to provide the CLI option alternative to enable flakes?

berewt (approved):
You can either address the comments or ignore them, up to you.

marconi-sidechain/README.md
neilmayhew: I see! So we can't build Marconi for `aarch64-darwin`? I didn't realize that.

I'll add an additional paragraph saying that "for now" we can't build native binaries for Apple Silicon and so x86 has to be forced using that arg.

marconi-sidechain/README.md
neilmayhew: I see no reason not to enable `nix-command flakes` on every nix system, so I think it's simpler just to tell people to do that. If someone really cares about not enabling them they're almost certainly a nix expert and will know about the CLI option alternative already.

marconi-sidechain/README.md
berewt: Yes, the native silicon doesn't build (I don't remember for which reason). I know the devx team is supposed to work on it.

marconi-sidechain/README.md
berewt: True, I was just thinking about the test via a docker image scenario, where it may sound less cumbersome to add a parameter rather than modify a config file. I agree though that the difference is minor and omitting it is probably less confusing.

marconi-sidechain/README.md
neilmayhew: I assume that means the Nix+Cabal option doesn't work either, so I've added notes to both Nix and Nix+Cabal. LMK if Nix+Cabal is OK and I'll take out that note.

cabal.project
```diff
@@ -20,10 +20,12 @@ packages: legacy/marconi-core-legacy
           marconi-chain-index
           marconi-core
           marconi-core-json-rpc
-          doc/read-the-docs-site
           marconi-sidechain
           marconi-sidechain-experimental
           marconi-starter
+if(!os(darwin) || impl(ghc >= 9.4))
```
sjoerdvisscher, line 26: Is it known that everything works fine with ghc 9.6?

neilmayhew: I haven't tried that yet but I thought it would be worth re-enabling it automatically when we do upgrade. If it's permanently disabled, no-one will remember to check whether we can re-enable this. If we find we get failures again we can just increase the bounds here.

PR #244, opened 2023-11-22
PLT-6108 - Add a CLI parameter to marconi-sidechain to make resume explicit

willjgould:
I've added three params:

* `start-from` - a command that takes `--slot-no` and `--block-header-hash`
* `start-from-last-sync-points`
* `start-from-genesis`

I've also run the tests to update the golden files.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


marconi-chain-index/src/Marconi/ChainIndex/CLI.hs
```diff
@@ -42,38 +42,82 @@ import Marconi.ChainIndex.Types (
...
+        <$> Opt.subparser
+          ( Opt.command
+              "start-from"
+              ( Opt.info
+                  ( C.ChainPoint
+                      <$> slotNoParser
+                      <*> blockHeaderHashParser
+                  )
+                  ( Opt.progDesc
+                      "Start from a given slot and block header hash. Useage: `start-from --slot-no SLOT_NO --block-header-hash BLOCK_HEADER_HASH`"
```
koslambrou, line 91: Useage -> Usage

marconi-chain-index/src/Marconi/ChainIndex/CLI.hs
```diff
@@ -126,7 +170,7 @@ data CommonOptions = CommonOptions
   -- ^ POSIX socket file to communicate with cardano node
   , optionsNetworkId :: !NetworkId
   -- ^ cardano network id
-  , optionsChainPoint :: !ChainPoint
+  , optionsChainPoint :: !StartFrom
   -- ^ Required depth of a block before it is indexed
```
koslambrou, line 174: Wrong comment

marconi-chain-index/src/Marconi/ChainIndex/CLI.hs
```diff
@@ -126,7 +170,7 @@ data CommonOptions = CommonOptions
   -- ^ POSIX socket file to communicate with cardano node
   , optionsNetworkId :: !NetworkId
   -- ^ cardano network id
-  , optionsChainPoint :: !ChainPoint
+  , optionsChainPoint :: !StartFrom
```
koslambrou, line 173: Rename to `StartingPoint`?

marconi-chain-index/test/Spec/Golden/CLIInputValidation/invalidAddress-MissingBech32Prefix.golden
```diff
@@ -2,8 +2,8 @@ option --addresses-to-index: Invalid address (not a valid Bech32 address represe
 
 Usage: marconi-chain-index [--version] (-s|--socket-path FILE-PATH) 
                            (--mainnet | --testnet-magic NATURAL) 
-                           [(-n|--slot-no SLOT-NO)
-                             (-b|--block-header-hash BLOCK-HEADER-HASH)] 
+                           [--start-from-genesis | 
+                             --start-from-last-sync-points | COMMAND]
```
koslambrou, line 6: What's that `COMMAND`? Can we put the actual expected data like `(--slotNo SLOTNO --block-header-hash BHH)`

marconi-chain-index/src/Marconi/ChainIndex/CLI.hs
```diff
@@ -42,38 +42,82 @@ import Marconi.ChainIndex.Types (
  )
 import Paths_marconi_chain_index (version)
 
-{- | Allow the user to set a starting point for indexing the user needs to provide both
- a @BlockHeaderHash@ (encoded in RawBytesHex) and a @SlotNo@ (a natural number).
+-- | Represents a specified point from which to start indexing
+data StartFrom
+  = StartFromGenesis
```
koslambrou, line 47: I'm not 100% sure here, but I'm wondering if we start from genesis, should we delete the marconi database files? In the case where resuming is enabled, the indexer will not reinitialise it's database. Instead, it will drain incoming event if it has already indexed them.

Probably a new ticket for that, but worth discussing it here.

cc @berewt

marconi-chain-index/test/Spec/Golden/CLIInputValidation/invalidAddress-MissingBech32Prefix.golden
willjgould: Yep, but it looks like I have to do it manually with `metavar`.

marconi-chain-index/src/Marconi/ChainIndex/CLI.hs
```diff
@@ -42,38 +42,84 @@ import Marconi.ChainIndex.Types (
...
-    maybeParseHashBlockHeader :: String -> Maybe (C.Hash C.BlockHeader)
-    maybeParseHashBlockHeader =
-      either (const Nothing) Just
-        . C.deserialiseFromRawBytesHex (C.proxyToAsType Proxy)
-        . C8.pack
+    givenPoint :: Opt.Parser StartingPoint
+    givenPoint =
+      StartFrom
+        <$> Opt.hsubparser
+          ( Opt.command
```
koslambrou, line 83: But why a command? I feel like you could omit the `start-from`, and have instead `--start-from (--slotNo ... --block-header-hash ...)`. Unless optparse doesn't permit that.

willjgould: We can, but it lets you do weird stuff like this (not that anyone actually would, hopefully):

`--start-from --node-config-path ~/git/marconi/config/cardano-node/preprod/config.json --slot-no 1 --block-header-hash asdasd`

marconi-chain-index/src/Marconi/ChainIndex/CLI.hs
willjgould: I don't think it should just delete with this flag, I think another flag should be required for that. It's a potential nasty sharp edge for users otherwise; someone might just want to try out the command without expecting it to wipe their DB.

marconi-chain-index/src/Marconi/ChainIndex/CLI.hs
berewt: `--start-from slotNo:blockHeaderHash`? Having slot number alone and not being able to process it is misleading anyway.

marconi-chain-index/src/Marconi/ChainIndex/CLI.hs
berewt: You may want to start again from a given point without wiping out the database. It's not obvious for genesis, but as it would be reasonable for other points, having the same behaviour for genesis is consistent.

marconi-chain-index/test/Spec/Golden/CLIInputValidation/invalidAddress-MissingBech32Prefix.golden
berewt: the `| COMMAND` part is confusing, it sounds like I can use `--start-from-genesis`, `--start-from-last-sync-points` or ANY command.

marconi-chain-index/src/Marconi/ChainIndex/CLI.hs
koslambrou: Good points. Also, having an additional option such as `--delete-database` makes sense. That should be part of a separate PR.

koslambrou (approved):

PR #259, opened 2023-12-01
PLT-7982 - streaming events

willjgould:
Added a `WithAction` class, which executes arbitrary IO on events during indexing, a `Streamable` typeclass, which allows us to describe things we can stream to and from, and used these constructs to create a function `withStream`, that gives us a streaming indexer.

Pre-submit checklist:
- Branch
    - [x] Important changes are reflected in changelog.d of the affected packages
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested


marconi-core/src/Marconi/Core/Transformer/WithStream/TChan.hs
```diff
@@ -0,0 +1,37 @@
...
+   . (MonadIO m)
+  => TChan r
+  -> Stream (Of r) m ()
+streamFromTChan chan = repeatM . liftIO . atomically $ readTChan chan
+
+-- | A smart constructor for @WithAction@
+withStream
+  :: (Show event, Show (Point event))
+  => (Timed (Point event) event -> r)
+  -> TChan r
```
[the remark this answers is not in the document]
willjgould: You're right, I'll switch to a `TBQueue`.

marconi-core/test/Marconi/CoreSpec.hs
```diff
@@ -1636,6 +1706,74 @@ withRollbackFailureTest =
...
+            [ Tasty.testProperty "TBQueue streaming works" withStreamTBQueue
+            ]
+        , Tasty.testGroup
+            "withStreamSocket"
+            [ Tasty.testProperty "Socket streaming works" withStreamSocket
+            ]
+        ]
+    ]
+
+withStreamTBQueue :: Property
```
koslambrou, line 1728: Property tests should start with `prop`. Also add a comment for each explaining _what_ your testing exactly.

koslambrou (commented):
After taking a quick look, I got a few comments:

* for any new constructs and modules, like `WithAction`, `TBQueue` and `Socket`, I would really to see some top level module documentation explaining what capability the interface provides, and when it should be used, along with maybe some examples.
* why is the `Socket` module necessary? Will we actually need it?
* is there an issue with the TBQueue stream where if there is no consuming client, then the indexer simply blocks once the queue is full?

In general, the idea seems good. Once we _actually_ use it for the LedgerEvents, we can probably identify more clearly if that design works or not.

willjgould:
@koslambrou 

> for any new constructs and modules, like WithAction, TBQueue and Socket, I would really to see some top level module documentation explaining what capability the interface provides, and when it should be used, along with maybe some examples.

Planning on it, just hadn't quite got round to it yet.

> why is the Socket module necessary? Will we actually need it?

I think we should chat about this tomorrow. I'm not sure, is the answer, but maybe.

> is there an issue with the TBQueue stream where if there is no consuming client, then the indexer simply blocks once the queue is full?

Good question, there probably is. I'll double check tomorrow and figure out a solution.

berewt:
>> is there an issue with the TBQueue stream where if there is no consuming client, then the indexer simply blocks once the queue is full?

> Good question, there probably is. I'll double check tomorrow and figure out a solution.

It will. There isn't any workaround, it's the user's responsibility. Other solutions are: to let the memory leak or lose events. It looks worse to me.

marconi-core/src/Marconi/Core/Transformer/WithAction.hs
```diff
@@ -0,0 +1,85 @@
...
+{- |
+  A transformer that fans events into an arbitrary action.
+
+ The configuration is a function @Timed (Point event) event -> IO ()@
+
+ See Marconi.Core.Transformer.WithStream.Socket and Marconi.Core.Transformer.WithStream.TBQueue
+ for example usages.
+
+ Once we are close enough to the tip of the chain, `WithCatchup` deactivate itself and pass directly
+ the blocks to the indexer.
```
koslambrou, line 13: I don't understand this comment. What does that have to do with `WithAction`?

marconi-core/src/Marconi/Core/Transformer/WithAction.hs
```diff
@@ -0,0 +1,85 @@
...
+
+ When you use several workers and coordinators, you may want to put the catchup on the final
+ indexers and not on the coordinators to improve performances.
```
koslambrou, line 16: Same comment here.

marconi-core/src/Marconi/Core/Transformer/WithStream/Streamable.hs
```diff
@@ -0,0 +1,133 @@
...
+import Network.Socket.ByteString.Lazy (sendAll)
+import Streaming (Of, Stream)
+import Streaming.Prelude (repeatM, yield)
+
+-- | Something that can have mapped events streamed to or from it
+class Streamable a r where
+  streamFrom :: a -> Stream (Of r) IO ()
+  streamTo :: (Timed (Point event) event -> r) -> a -> Timed (Point event) event -> IO ()
+
+{- | A streaming implementation using @Socket@. Here's an example of how
```
koslambrou, line 21: If that's already implemented in the test module, I would just point the user to that test code in the test module.

marconi-core/src/Marconi/Core/Transformer/WithStream/Streamable.hs
```diff
@@ -0,0 +1,133 @@
...
+  streamTo mapping sock = sendAll sock . encode . mapping
+
+{- | An experimental streaming implementation using @TBQueue@.
+
+    IMPORTANT: Note that this will block when used with @withStream@ if you have slow or
+               non-existent consumers!
+
+    Here's an example of how to use it, in the form of a property test:
+
+    propWithStreamTBQueue :: Property
```
koslambrou, line 98: Same comment here as above.

koslambrou (approved):
LGTM. Minor comments.

marconi-core/src/Marconi/Core/Transformer/WithAction.hs
willjgould: Ah my bad, I copy and pasted the structure of `WithCatchup`'s description and must have forgotten to remove the rest!

PR #266, opened 2023-12-11
PLT-8634 test structure marconi-sidechain-experimental

brendanrbrown:
Creates a base on which we can add tests of handlers to `marconi-sidechain-experimental`. The general approach is to generate events for a Mockchain and to test JSON RPC query handlers directly, without needing to run a node or the HTTP server.

* Extracts various generators and utilities for creating events from `marconi-cardano-indexers-test`, making them available in `marconi-cardano-indexers-test-lib`.
* Adds generators for the events needed to test handlers to the new modules `marconi-cardano-indexers-test-lib`.
   * These mainly are transformations to use the existing Mockchain generators and create the types needed for indexing events with the types needed for queries tested here.
* Adds a derived `Show` instance for `WithDistance` so these events can be generated with hedgehog.
* Creates module `Test.Gen.Marconi.Cardano.Indexers` with a version of `buildIndexer` that exposes the underlying indexers in the coordinator used for `SidechainHttpConfig`.
   * Needed (for now) so that generated events can be indexed directly without needing to pass through the chain-sync protocol, whose 'BlockEvent' type we cannot generate.
   * The module also includes utilities for directly indexing events to the indexers used for tests in this PR.
* Provides tests of basic query results for the `getUtxosFromAddress` and `getBurnTokenEvents`.
* Sets `CARDANO_NODE_CONFIG` to make the cardano node `config` directory available both in CI checks and the nix shell.

Note:

* Utilities for directly indexing events only cover the needs of the handler tests added here. Future PRs can just extend the existing utilities to support indexing other event types as needed for testing additional handlers.

Pre-submit checklist:
- Branch
    - [X] Tests are provided (if possible)
    - [X] Commit sequence broadly makes sense and have useful messages
    - [ ] Important changes are reflected in changelog.d of the affected packages
    - [X] Relevant tickets are mentioned in commit messages
- PR
    - [ ] (For external contributions) Corresponding issue exists and is linked in the description
    - [X] Targeting main unless this is a cherry-pick backport
    - [X] Self-reviewed the diff
    - [X] Useful pull request description
    - [ ] If relevant, reference the ADR in the PR and reference the PR in the ADR
    - [x] Reviewer requested

PR #268, opened 2023-12-12
Remove unused env var from marconi-chain-index.preCheck

neilmayhew:
This also removes a nix dependency on the source package

The var was being used by some commented-out tests that needed access to the node config. In the unlikely event that those tests are reinstated (they're more likely to be migrated to using the node emulator) we can have them use the solution that Brendan has just introduced for his recent PR (#266).


brendanrbrown (approved):
This seems reasonable to me, along with your suggested fix should we need to add it back in. Thanks Neil.

PR #277, opened 2023-12-14
Update GitHub repo URLs for the IntersectMBO migration

neilmayhew:
I don't know if the PR and Issue URLs will carry across as well


berewt:
It's premature, we have:

1. to check the list of migrated repo
2. to check with IOG how they want to proceed with the update

PR #282, opened 2023-12-21
Add DBSync Comparison testing framework

ana-pantilie:
### Changes implemented in this PR
- Adds the possibility of running indexers on mainnet snapshots, generalises and refactors into multiple modules.
- Adds 27 `BlockInfo` Db-Sync Comparison tests.
- Adds documentation on how to add new tests.
- In order to run the tests with an in-memory `marconi` database, I implemented [5465376](https://github.com/input-output-hk/marconi/pull/282/commits/5465376d963a08892e301f923d458889eb7058c0). It is not perfect, so there might be a bit more work to be done here.

### Very important notes:
- The tests will fail on CI because the snapshots are not yet on CI. **We need https://input-output.atlassian.net/browse/PLT-9092?atlOrigin=eyJpIjoiZDEzNmUxYjgyNzUxNDE1MThlNDViYjYyYTYwYjc2OTciLCJwIjoiaiJ9 in order to continue** -> used `ignoreTest` so that the test suite runner ignores them
- The added tests are failing because Db-Sync outputs an absolute time as the `time` part of the `BlockInfo`. My guess it that `marconi` seems to be keeping the time difference between when the current block was emitted vs. the time at which the block before it was. I don't know which one of these results is expected, so more discussion is needed. Depending on that we can either modify the `golden` files or modify `marconi` to better conform to Db-Sync's style of output. -> opened separate issue about this
- ~You will notice that every test for the slot at the end of the snapshot returns `null`. This is most likely a bug in the testing framework which I didn't have time to fix.~ fixed [016e413](https://github.com/input-output-hk/marconi/pull/282/commits/016e4132e58ac1876fd5427731ce6ba849ffa744)

Pre-submit checklist:
- Branch
- PR
    - [ ] Self-reviewed the diff
    - [ ] Reviewer requested


marconi-core/src/Marconi/Core/Indexer/SQLiteAggregateQuery.hs
```diff
@@ -83,7 +83,7 @@ mkSQLiteAggregateQuery
 mkSQLiteAggregateQuery sources = do
   con <- liftIO $ SQL.open ":memory:"
   let databaseFromSource :: SQLiteSourceProvider m point -> IO FilePath
-      databaseFromSource (SQLiteSourceProvider ix) = Con.withMVar ix (pure . getDatabasePath)
+      databaseFromSource (SQLiteSourceProvider ix) = Con.withMVar ix (pure . SQLite.extractStorageUnsafe . getDatabasePath)
```
berewt, line 86: Remove `extractStorageUnsafe`, and explicitly throw an `Exception` if one of the source is an in-memory db.

berewt (commented):
If we remove the partial function we should be ok :)

berewt (approved):

PR #291, opened 2024-01-14
PLT-9261 possible fix and debug for intermittent darwin failure

brendanrbrown:
* Use a non-empty temp directory prefix. This is a guess at a fix for a somewhat rare failure on darwin, from the socket path not being found in `propWithStreamSocket` from `marconi-core-test`.
* Reports the socket path on failure -- swallowing the original exception and failing the test after reporting the socket path. See the code note on this workaround.

The second item is to help find a real fix for this intermittent problem if needed. See PLT-9260.

Pre-submit checklist:
- Branch
    - [ ] Tests are provided (if possible)
    - [ ] Commit sequence broadly makes sense and have useful messages
    - [ ] Important changes are reflected in changelog.d of the affected packages
    - [X] Relevant tickets are mentioned in commit messages
- PR
    - [ ] (For external contributions) Corresponding issue exists and is linked in the description
    - [X] Targeting main unless this is a cherry-pick backport
    - [X] Self-reviewed the diff
    - [X] Useful pull request description
    - [ ] If relevant, reference the ADR in the PR and reference the PR in the ADR
    - [X] Reviewer requested

PR #293, opened 2024-01-16
PLT-9276 Properly disconnect catchup when we reach the expected distance

berewt:
It introduces two constructor for `WithCatchup` to differentiate from the datatype when we're catching up and when we're done with it.

Pre-submit checklist:
- Branch
- PR
    - [x] Self-reviewed the diff
    - [ x] Reviewer requested

### DIFF
--- a/marconi-core/src/Marconi/Core/Transformer/WithCatchup.hs
+++ b/marconi-core/src/Marconi/Core/Transformer/WithCatchup.hs
@@ -22,9 +22,10 @@ module Marconi.Core.Transformer.WithCatchup (
 ) where
 
 import Control.Lens qualified as Lens
-import Control.Lens.Operators ((%~), (+~), (.~), (?~), (^.))
+import Control.Lens.Operators ((%~), (+~), (.~), (^.), (^?))
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Function ((&))
+import Data.Maybe (fromMaybe)
 import Data.Word (Word64)
 import Marconi.Core.Class (
   Closeable,
@@ -69,7 +70,7 @@ mkCatchupConfig batchSize bypassDistance = CatchupConfig batchSize bypassDistanc
 
 Lens.makeLenses ''CatchupConfig
 
-data CatchupContext event = CatchupContext
+data Catchingup event = Catchingup
   { _contextDistanceComputation :: Point event -> event -> Word64
   -- ^ How we compute distance to tip
   , _contextCatchupConfig :: CatchupConfig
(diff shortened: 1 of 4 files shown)


marconi-core/src/Marconi/Core/Transformer/WithCatchup.hs
```diff
@@ -191,48 +196,58 @@ instance
...
+              Nothing -> pure ix'
+              Just lastStable' -> setLastStablePointVia caughtUpIndexer lastStable' ix'
+            pure $ resetBuffer ix''
+          triggerHook = case cfg ^. contextCatchupEventHook of
+            Nothing -> pure ()
+            Just hook -> liftIO $ hook Synced
+      if hasCaughtUp
+        then do
+          triggerHook
+          indexer' <- sendBatch indexer
```
brendanrbrown, line 230: `sendBatch` in some cases calls `setLastStablePointVia`. In the previous version of this section, the last stable point was not updated when the catchup buffer was empty: The original indexer was returned.

Was that intended? From what I see I think it is a small but inoffensive difference: The last stable point might be updated more frequently.

The remaining stuff in `sendBatch` is a no-op when the buffer is empty, I think.

brendanrbrown (approved):
This makes sense to me. I made one comment that is worth a closer look: See the `sendBatch` note. I think it's fine as-is, but it would be good to double-check.

Remaining comments are very minor.

brendanrbrown:
Oh, and I see the temporary fix in #291 did not fix the darwin socket file intermittent errors. Looking into that now, but it's fine by me if we want to re-run the test in the meantime and hope for better.

marconi-core/src/Marconi/Core/Transformer/WithCatchup.hs
berewt: It may call it one time more. `hasCaughtUp` should be true only one time during synchronisation: when we reach the expected distance to the tip.

marconi-core/src/Marconi/Core/Transformer/WithCatchup/SQLite.hs
```diff
@@ -71,3 +69,8 @@ createIndexTable indexerName stdoutTrace c indexName createIndexStatement = do
         <> "' for "
         <> indexerName
         <> " indexer..."
+
+autoVacuum :: Trace IO Text -> SQL.Connection -> IO ()
+autoVacuum logger c = do
+  SQL.execute_ c "PRAGMA auto_vaccum = 1"
```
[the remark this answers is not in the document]
berewt: It was the initial use case for it, I wanted to set auto vacuum on the indexers but after reading the doc it looks like in our case it wouldn't, change anything, so we can actually remove it.

